PR #14, opened 2020-10-01
Exclude statuses except Published

brinxmat:
Fixes the bug that unpublished publications are indexed when they should not be.


svenngjeto (approved):
No comment, plain and nice

axthosarouris (approved):

PR #21, opened 2020-11-02
Np 1164 add publisher

svenngjeto:


search-resources-api/src/main/java/no/unit/nva/dynamodb/IndexDocumentGenerator.java
```diff
@@ -118,6 +123,13 @@ private static URI extractDoi(JsonNode record) {
                 .orElseThrow();
     }
 
+    private static URI extractPublisherId(JsonNode record) {
+        return Optional.ofNullable(record)
```
brinxmat, line 127: It occurs to me that the id of publisher should be a URI, as it stands, it is a document-internal node, which is unproblematic, but I think we will revisit this in a later PR.

search-resources-api/src/main/java/no/unit/nva/search/IndexPublisher.java
```diff
@@ -0,0 +1,78 @@
...
+import java.net.URI;
+import java.util.Objects;
+
+public class IndexPublisher {
+    private final URI id;
+    private final String name;
+
+    @JacocoGenerated
+    @JsonCreator
+    public IndexPublisher(@JsonProperty("id") URI id,
```
brinxmat, line 16: As I said in the previous comment, we need to look at the publishers API (I don't think it exists)…I think we have customer API, which is probably the wrong thing…we should have a separate API for publishers (yes, served from the customer table

brinxmat (approved):

PR #24, opened 2020-11-04
Np 1164 list publications with sorting on modifiedDate desc

svenngjeto:


template.yaml
```diff
@@ -102,8 +101,33 @@ Resources:
         openapi: 3.0.3
         info:
           title: NVA Search API
-          version: '0.0.2'
+          version: '0.1.0'
         paths:
+          /:
```
brinxmat, line 106: Is this actually updated with params?

svenngjeto: OpenAåpi is updated, but it doesn't show within the current setup - this is for all our api's defined in openapi..

README.md
```diff
@@ -1,31 +1,88 @@
...
+**Parameters to GET /search/resources**
 
-Structure wanted in response from the search API
+|Parameter|Description|Default value |
+|-----|-----|-----|
+| query | Term to search for in resource. | * 
+| orderBy | Field to order result by. |  modifiedDate 
+| sortOrder | Order of search results (**asc**ending or **desc**ending). | desc
+| from | Start posision for results, 0-based. | 0
+| results | Maximum number of publications in response. | 10
```
brinxmat, line 12: Is there a max results? I.e. 10 (max 100)?

README.md
```diff
@@ -1,31 +1,88 @@
...
+      ],
+      "title": "Toward unique identifiers",
+      "owner": "tehe@unit.no",
+      "publishedDate": {
+        "type": "IndexDate",
+        "year": "1999"
+      },
+      "publisher": {
+        "id": "https://api.dev.nva.aws.unit.no/customer/f54c8aa9-073a-46a1-8f7c-dde66c853934",
+        "name": "Organization"
```
brinxmat, line 80: Name Organization indicates, maybe we have an issue

brinxmat (approved):

README.md
svenngjeto: Max results in this particular response, elasticSearch uses int as type for both from and results - for pagination. If one is interesteded in emptying the search database '_scroll' comes in handy....

README.md
svenngjeto: This is an issue as we don't enriche the data besides what is in the event from DynamoDb. Maybe we must enrich search index with data from multiple sources?

PR #43, opened 2021-03-15
NP-2318: DynamoDBStreamHandler extends EventHandler

axthosarouris:


build.gradle
```diff
@@ -107,15 +107,15 @@ subprojects {
                 limit {
                     counter = 'METHOD'
                     value = 'COVEREDRATIO'
-                    minimum = 1.00
+                    minimum = 0.00
```
axthosarouris, line 110: All tests for DynamoDBStreamHandler have been disabled. I will fix one test per PR

brinxmat (approved):

PR #49, opened 2021-03-16
tests for deleting and unknown operation

axthosarouris:


build.gradle
```diff
@@ -61,6 +61,8 @@ subprojects {
     dependencies {
 
         testImplementation project(":test-commons")
+        testImplementation group: 'com.github.javafaker', name: 'javafaker', version: '1.0.2'
```
axthosarouris, line 64: No reason testing libraries not to be common for all modules

brinxmat (approved):

PR #55, opened 2021-03-17
Np 2318 incomplete publication dates

axthosarouris:


search-resources-dynamodb-trigger/src/test/java/no/unit/nva/publication/DynamoDBStreamHandlerTest.java
```diff
@@ -369,30 +369,7 @@ private DeleteResponse mockDeleteResponse() {
     }
 
 
-    //
-    //
```
axthosarouris, line 373: Test was deleted, because we do not do anything specific with incomplete dates. Right now we do as before, namely we copy the year month and date. This is tested in previous tests.

svenngjeto (approved):

brinxmat (approved):

torarnet (approved):

PR #56, opened 2021-03-18
Np 2318 empty contributor list

axthosarouris:


search-resources-dynamodb-trigger/src/test/java/no/unit/nva/publication/DynamoDBStreamHandlerTest.java
```diff
@@ -369,48 +369,6 @@ private DeleteResponse mockDeleteResponse() {
     }
 
 
-
```
axthosarouris, line 372: These tests we replaced by tests in IndexDocumentTest

brinxmat (approved):

PR #63, opened 2021-04-14
bugfix: re-enable resource-deletion

axthosarouris:


template.yaml
```diff
@@ -556,5 +556,52 @@ Resources:
             Type: SQS
             Destination: !GetAtt ResourcesIndexerDLQ.Arn
 
+  DeleteResources:
```
axthosarouris, line 559: Create a copy of the IndexResources handler, but with different event pattern to match the deletion of a resource

brinxmat (approved):

template.yaml
svenngjeto: Maybe we should always try to delete a resource from searchIndex, independent of it is published or not - if the resource is deleted from the 'master' database.

PR #74, opened 2021-06-21
NP-2801: use  index instead of update.

axthosarouris:


search-commons/src/main/java/no/unit/nva/search/ElasticSearchHighLevelRestClient.java
```diff
@@ -179,15 +177,15 @@ private SearchRequest getSearchRequest(String term, int results, int from, Strin
     }
 
     private void doUpsert(IndexDocument document) throws IOException {
```
torarnet, line 179: doUpsert?

torarnet (approved):

svenngjeto (approved):

PR #89, opened 2021-07-05
bugfix: make es smaller

axthosarouris:


svenngjeto (approved):
We have tied used t2.small before, but with the changed import it might work better.. It's half the price...

PR #99, opened 2021-08-24
Framing example

brinxmat:
This is an example, don't merge the code

It shows how to frame the input to a coherent document.


brinxmat:
Example code, removed

PR #104, opened 2021-09-22
Separate batch indexing logic from lambda handler

axthosarouris:


batch-index/src/main/java/no/unit/nva/search/BatchIndexer.java
```diff
@@ -0,0 +1,118 @@
...
+import no.unit.nva.s3.UnixPath;
+import no.unit.nva.search.exception.SearchException;
+import nva.commons.core.JsonUtils;
+import nva.commons.core.attempt.Try;
+import nva.commons.core.exceptions.ExceptionUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import software.amazon.awssdk.services.s3.S3Client;
+
+public class BatchIndexer {
```
axthosarouris, line 26: The code in this class was mainly a cut-paste from the handler.

brinxmat (dismissed):

PR #114, opened 2021-09-27
Improve bulk indexing

axthosarouris:


batch-index/src/main/java/no/unit/nva/search/PrepareIndexForBatchInsertHandler.java
```diff
@@ -1,41 +0,0 @@
-package no.unit.nva.search;
```
axthosarouris, line 1: This class is not used

batch-index/src/test/java/no/unit/nva/search/PrepareIndexForBatchInsertHandlerTest.java
```diff
@@ -1,149 +0,0 @@
-package no.unit.nva.search;
```
axthosarouris, line 1: Ditto

build.gradle
```diff
@@ -35,7 +35,7 @@ allprojects {
         nvaDatamodelVersion = '0.13.1'
         nvaTestUtilsVersion = "0.1.22"
         testContainersVersion = '1.15.2'
-        nvaPublicationApiVersion = '0.3.2'
+        nvaPublicationApiVersion = '0.3.3'
```
axthosarouris, line 38: Using the newest IonReader from nvaPublication service. This library will eventually be moved to commons if we continue to use the IonFormat. However, in the future the re-indexing process will happen from an export that is produced automatically using DynamoDbStreams and contains all resources in JsonFormat. In such case, the IonReader will not be necessary.

search-commons/src/main/java/no/unit/nva/search/ElasticSearchHighLevelRestClient.java
```diff
@@ -151,6 +153,7 @@ public AcknowledgedResponse prepareIndexForBatchInsert() throws IOException {
     public List<BulkResponse> batchInsert(List<IndexDocument> indexDocuments) {
         List<List<IndexDocument>> bulks = Lists.partition(indexDocuments, BULK_SIZE);
         return bulks.stream()
+            .parallel()
```
axthosarouris, line 156: sending bulk indexing requests in parallel, speeds up indexing a great deal.

template.yaml
```diff
@@ -136,10 +136,11 @@ Resources:
             Resource: !Sub 'arn:aws:es:${AWS::Region}:${AWS::AccountId}:domain/*'
       ElasticsearchClusterConfig:
         DedicatedMasterEnabled: true
-        DedicatedMasterType: t2.small.elasticsearch
-        InstanceCount: 1
+        DedicatedMasterType: t2.medium.elasticsearch
```
axthosarouris, line 139: this works fine for now. We can test with small again after we have something working in DEV

brinxmat (approved):

janbj (approved):

PR #118, opened 2021-10-01
fix memory leak

axthosarouris:


batch-index/src/main/java/no/unit/nva/search/S3IonReader.java
```diff
@@ -0,0 +1,57 @@
+package no.unit.nva.search;
```
axthosarouris, line 1: This class was transfered from publication-api and modified for better memory usage. The next step will be to be transfered to commons and used by publication-api as well.

axthosarouris: It is tested indirectly, through the tests of the batch indexer

batch-index/src/test/java/no/unit/nva/search/BatchIndexTest.java
```diff
@@ -34,15 +35,15 @@
     protected StubElasticSearchHighLevelRestClient failingElasticSearchClient() {
         return new StubElasticSearchHighLevelRestClient() {
             @Override
-            public List<BulkResponse> batchInsert(List<IndexDocument> indexDocuments) {
-                List<BulkItemResponse> itemResponses = indexDocuments.stream()
+            public Stream<BulkResponse> batchInsert(Stream<IndexDocument> indexDocuments) {
```
axthosarouris, line 38: changing the signature of overriding method.

search-commons/src/main/java/no/unit/nva/search/ElasticSearchHighLevelRestClient.java
```diff
@@ -152,23 +154,10 @@ public AcknowledgedResponse prepareIndexForBatchInsert() throws IOException {
...
-
-    private BulkResponse insertBatch(List<IndexDocument> bulk) throws IOException {
-        BulkRequest request = new BulkRequest();
-        for (IndexDocument document : bulk) {
-            request.add(getUpdateRequest(document));
-        }
-        request.setRefreshPolicy(RefreshPolicy.WAIT_UNTIL);
-        request.waitForActiveShards(ActiveShardCount.ONE);
-        return elasticSearchClient.bulk(request, RequestOptions.DEFAULT);
+    public Stream<BulkResponse> batchInsert(Stream<IndexDocument> indexDocuments) {
```
axthosarouris, line 157: replaced list with stream so that we do not have to load all index documents at once.

brinxmat (approved):

janbj (approved):

PR #127, opened 2021-10-07
Np3243 illegal query

svenngjeto:
URLEncoding incoming query before sending it to ElasticSearch


gradle.properties
```diff
@@ -1,3 +1,2 @@
 org.gradle.parallel=true
 org.gradle.caching=true
-org.gradle.console=verbose
```
brinxmat, line 3: Do we want no verbose output by default?

brinxmat (approved):

PR #165, opened 2021-11-30
Np3021 refactor search client constructor

janbj:


search-commons/src/main/java/no/unit/nva/search/SearchClientConfig.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    public static SearchClient defaultSearchClient() {
+        return new SearchClient(defaultRestHighLevelClientWrapper());
+    }
+
+    public static RestHighLevelClientWrapper defaultRestHighLevelClientWrapper() {
+        return createElasticsearchClientWithInterceptor(
+                ELASTICSEARCH_ENDPOINT_ADDRESS,
+                ELASTICSEARCH_ENDPOINT_INDEX
```
axthosarouris, line 37: why do you need the index here? only for logging?

janbj: Removed index argument

search-commons/src/main/java/no/unit/nva/search/SearchClient.java
```diff
@@ -136,25 +87,26 @@ private SearchResponse doSearch(String term,
...
-    private SearchRequest getSearchRequest(String term, int results, int from, String orderBy, SortOrder sortOrder) {
+    private SearchRequest getSearchRequest(
+            String searchTerm, int results, int from, String orderBy, SortOrder sortOrder) {
         final SearchSourceBuilder sourceBuilder = new SearchSourceBuilder()
-                .query(QueryBuilders.queryStringQuery(term))
+                .query(QueryBuilders.queryStringQuery(searchTerm))
                 .sort(orderBy, sortOrder)
                 .from(from)
                 .size(results);
         return new SearchRequest(ELASTICSEARCH_ENDPOINT_INDEX).source(sourceBuilder);
```
axthosarouris, line 97: I assume that this will change in later PR and be dynamic, so that we can serve all indexes?

janbj: Hm, good point. I can do it now.

axthosarouris (approved):

PR #184, opened 2022-01-06
Np 3522 add viewing scope parameter to search

axthosarouris:
add a query parameter to the endpoint so that the following query is possible:
```
https://api.dev.nva.aws.unit.no/search/doirequests/search?viewingScope=https://api.dev.nva.aws.unit.no/cristin/organization/194.0.0.0
```

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -17,40 +17,18 @@ paths:
       parameters:
         - in: path
           name: index
-          description: "Index to search for resource in."
+          description: "Index to search for resources in."
           required: true
           schema:
             type: string
         - in: query
-          name: query
-          description: "Term to search for in resource."
+          name: viewingScope
+          description: "List resources for this organization."
           required: false
           schema:
             type: string
-        - in: query
-          name: orderBy
-          description: "Field to order result by."
-          required: false
-          schema:
-            type: string
-        - in: query
-          name: sortOrder
-          description: "Order of search results (asc or desc)."
-          required: false
-          schema:
-            type: string
-        - in: query
-          name: from
-          description: "Start position from results, 0-based."
-          required: false
-          schema:
-            type: string
-        - in: query
-          name: results
-          description: "Maximum number of resources in response."
-          required: false
-          schema:
-            type: string
+            format: uri
+            example: "https://api.dev.nva.aws.unit.no/cristin/organization/194.0.0.0"
       security:
         - CognitoUserPool: [ ]
       x-amazon-apigateway-integration:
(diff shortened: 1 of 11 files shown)


docs/openapi.yaml
```diff
@@ -17,40 +17,18 @@ paths:
...
-        - in: query
-          name: from
-          description: "Start position from results, 0-based."
-          required: false
-          schema:
-            type: string
-        - in: query
-          name: results
-          description: "Maximum number of resources in response."
+          name: viewingScope
```
axthosarouris, line 25: viewingScope is the only query parameter for now

search-commons/src/main/java/no/unit/nva/search/SearchClient.java
```diff
@@ -48,7 +46,7 @@ public SearchResourcesResponse searchSingleTerm(Query query, String index)
         return toSearchResourcesResponse(query.getRequestUri(), query.getSearchTerm(), searchResponse.toString());
     }
 
-    public SearchResponse findResourcesForOrganizationIds(String index, UserResponse.ViewingScope viewingScope)
+    public SearchResponse findResourcesForOrganizationIds(String index, ViewingScope viewingScope)
```
axthosarouris, line 49: ViewingScope is a separate class, since it is not always connected to a UserResponse

search-commons/src/test/java/no/unit/nva/search/restclients/responses/ViewingScopeTest.java
```diff
@@ -0,0 +1,16 @@
...
+
+import static no.unit.nva.testutils.RandomDataGenerator.randomUri;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.collection.IsIterableContainingInOrder.contains;
+import org.junit.jupiter.api.Test;
+
+public class ViewingScopeTest {
+
+    @Test
+    void shouldReturnNewViewingScopeWhenSupplyingUris() {
```
axthosarouris, line 11: The `create()` method is used in a different package and therefore a test is necessary here for the test coverage

docs/openapi.yaml
```diff
@@ -17,40 +17,18 @@ paths:
       parameters:
         - in: path
           name: index
-          description: "Index to search for resource in."
+          description: "Index to search for resources in."
```
brinxmat, line 20: I think this needs re-writing as it is not clear to me what this means, not because it is wrong, but the terms are all generic. I'd guess "Index for NVA-resources"

brinxmat (approved):

brinxmat (approved):

PR #186, opened 2022-01-07
Moving to gradle 7 first step

axthosarouris:


batch-index/build.gradle
```diff
@@ -2,8 +2,16 @@ dependencies {
     implementation project(':search-commons')
     implementation project(':indexing-handlers')
 
+    implementation group: 'com.github.bibsysdev', name: 'core', version: project.ext.nvaCommonsVersion
```
axthosarouris, line 5: Added all dependencies that before were declared in the commons as "api" dependencies

batch-index/src/main/java/no/unit/nva/search/BatchIndexingConstants.java
```diff
@@ -37,7 +36,7 @@ public static IndexingClient defaultEsClient() {
 
     @JacocoGenerated
     public static S3Client defaultS3Client() {
-        String awsRegion = ENVIRONMENT.readEnvOpt(AWS_REGION_ENV_VARIABLE).orElse(Regions.EU_WEST_1.getName());
+        String awsRegion = ENVIRONMENT.readEnvOpt(AWS_REGION_ENV_VARIABLE).orElse(Region.EU_WEST_1.toString());
```
axthosarouris, line 39: use aws sdk2 instead of sdk1

brinxmat (approved):

PR #209, opened 2022-05-19
Added new Lambda function for creating required initial indexes in ES

RanjithTom:


indexing-handlers/src/main/java/no/unit/nva/indexing/handlers/AllIndexHandler.java
```diff
@@ -0,0 +1,35 @@
+package no.unit.nva.indexing.handlers;
+
+import static nva.commons.core.attempt.Try.attempt;
+import com.amazonaws.services.lambda.runtime.Context;
+import com.amazonaws.services.lambda.runtime.RequestHandler;
+import no.unit.nva.search.IndexingClient;
+import nva.commons.core.JacocoGenerated;
+
+public class AllIndexHandler implements RequestHandler<Object, String> {
```
axthosarouris, line 9: IndexCreationHandler

indexing-handlers/src/test/java/no/unit/nva/indexing/handlers/AllIndexHandlerTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @BeforeEach
+    void init() {
+        indexingClient = mock(IndexingClient.class);
+        allIndexHandler = new AllIndexHandler(indexingClient);
+        context = mock(Context.class);
+    }
+
+    @Test
+    void shouldSentProperRequestToESToCreateIndexes() throws IOException {
```
axthosarouris, line 35: I am not sure you are testing what you are claiming here. How do we verify that the request is proper? The only thing we are testing is that the response we are getting from the handler is OK. 

If we are testing whether a request is proper, then we should get somehow this request and assert that its structure is the expected structure. We do not do this in this test. Here we are mocking the indicesClient and therefore we are basically accepting everything.  So what we are actually testing here is that the doNothing() method is actually not throwing an Exception.

I am not claiming that the test is wrong. I am not sure whether the name is in accordance with that is actually being tested. For me the test is more like "shouldNotThrowExceptionIfIndicesClientDoesNotThrowException".

axthosarouris (approved):
After merging, delete the nva-sandbox es stack and deploy again to check whether the code works

indexing-handlers/src/main/java/no/unit/nva/indexing/handlers/IndexCreationHandler.java
```diff
@@ -0,0 +1,35 @@
...
+        this(new IndexingClient());
+    }
+
+    public IndexCreationHandler(IndexingClient indexingClient) {
+        this.indexingClient = indexingClient;
+    }
+
+    @Override
+    public String handleRequest(Object input, Context context) {
+        var indicesClientWrapper = indexingClient.getIndicesClientWrapper();
```
axthosarouris, line 28: why do you need to expose the indicesClientWrapper and then send it back to the same obejct as parameter? There is no need for the indicesClientWrapper to be exposed.

indexing-handlers/src/main/java/no/unit/nva/indexing/handlers/IndexCreationHandler.java
```diff
@@ -0,0 +1,35 @@
...
+
+        attempt(() -> indexingClient.createIndexBasedOnName(RESOURCES, indicesClientWrapper)).orElseThrow();
```
axthosarouris, line 30: createIndex instead of createIndexBasedOnName. the "basedOnName" implies we have another method based on something else.

search-commons/src/main/java/no/unit/nva/search/IndexingClient.java
```diff
@@ -84,6 +85,26 @@ public void removeDocumentFromIndex(String identifier) throws IOException {
...
+    }
+
+    /**
+     * Provides an IndicesClientWrapper which can be used to access the Indices API. See Indices API on elastic.co
+     *
+     * @return IndicesClientWrapper.
+     */
+    public IndicesClientWrapper getIndicesClientWrapper() {
+        return elasticSearchClient.indices();
+    }
```
axthosarouris, line 106: you do not need to expose this to the clients of the class.

search-commons/src/test/java/no/unit/nva/search/IndexingClientTest.java
```diff
@@ -87,6 +91,36 @@ void shouldSendIndexRequestWithIndexNameSpecifiedByIndexDocument() throws IOExce
...
+    @Test
+    void shouldVerifyIndicesClientCreatedToAccessIndicesAPI() {
+        IndicesClientWrapper expectedIndicesClientWrapper = createMockIndicesClientWrapper();
+        when(esClient.indices()).thenReturn(expectedIndicesClientWrapper);
+        var actualIndicesClientWrapper = indexingClient.getIndicesClientWrapper();
+        assertEquals(expectedIndicesClientWrapper, actualIndicesClientWrapper);
+    }
+
+    @Test
+    void shouldVerifyProperIndexCreateRequestSentToIndicesClient() throws IOException {
```
axthosarouris, line 103: shouldCallEsClientCreateIndexRequest

search-commons/src/test/java/no/unit/nva/search/IndexingClientTest.java
```diff
@@ -87,6 +91,36 @@ void shouldSendIndexRequestWithIndexNameSpecifiedByIndexDocument() throws IOExce
         assertThat(extractDocumentFromSubmittedIndexRequest(), is(equalTo(indexDocument.getResource())));
     }
 
+    @Test
+    void shouldVerifyIndicesClientCreatedToAccessIndicesAPI() {
+        IndicesClientWrapper expectedIndicesClientWrapper = createMockIndicesClientWrapper();
+        when(esClient.indices()).thenReturn(expectedIndicesClientWrapper);
+        var actualIndicesClientWrapper = indexingClient.getIndicesClientWrapper();
+        assertEquals(expectedIndicesClientWrapper, actualIndicesClientWrapper);
+    }
```
axthosarouris, line 100: do we actually need this?

axthosarouris (changes requested):

indexing-handlers/src/main/java/no/unit/nva/indexing/handlers/IndexCreationHandler.java
RanjithTom: Thought of using the same indices client wrapper class instead of creating it each time when creating index, but as you said instead of exposing it to the client better to use it in indexingclient itself will update the code

search-commons/src/test/java/no/unit/nva/search/IndexingClientTest.java
RanjithTom: Going to remove getIndicesClientWrapper method then this test itself has no scope

search-commons/src/main/java/no/unit/nva/search/IndexingClient.java
```diff
@@ -84,11 +85,31 @@ public void removeDocumentFromIndex(String identifier) throws IOException {
         }
     }
 
+    /**
+     * Create Index in Elastic search based on the name.
+     *
+     * @param indexName name of the index needs to create.
+     */
+    public Void createIndex(String indexName) throws IOException {
```
axthosarouris, line 93:
```
   public Void createIndex(String indexName) throws IOException {
        elasticSearchClient.indices().create(new CreateIndexRequest(indexName), RequestOptions.DEFAULT);
        return null;
    }
```

search-commons/src/main/java/no/unit/nva/search/IndexingClient.java
```diff
@@ -84,11 +85,31 @@ public void removeDocumentFromIndex(String identifier) throws IOException {
         }
     }
 
+    /**
+     * Create Index in Elastic search based on the name.
```
axthosarouris, line 89: The comment does not provide information that we do know already. Remove it.

search-commons/src/main/java/no/unit/nva/search/IndexingClient.java
```diff
@@ -84,11 +85,31 @@ public void removeDocumentFromIndex(String identifier) throws IOException {
...
     }
 
+    /**
+     * Provides an IndicesClientWrapper which can be used to access the Indices API. See Indices API on elastic.co
+     *
+     * @return IndicesClientWrapper.
+     */
+    private IndicesClientWrapper getIndicesClientWrapper() {
+        return elasticSearchClient.indices();
+    }
```
axthosarouris, line 111: delete

axthosarouris (changes requested):

axthosarouris (approved):

brinxmat (approved):

PR #238, opened 2022-11-07
Np 16243/token

sondrev:


fredrikpf (approved):
Looks very good! 
Could have som tests on Interceptor and checking header?

fredrikpf (approved):

PR #254, opened 2023-02-06
Opensearch response to camelcase

fredrikpf:


search-commons/src/test/java/no/unit/nva/search/models/SearchResponseDtoTest.java
```diff
@@ -38,6 +49,24 @@ void jsonSerializationShouldKeepDeprecatedFieldsUntilFrontendHasMigrated() throw
...
+        var aggregations = SearchResponseDto.fromSearchResponse(searchResponse, randomUri()).getAggregations();
+
+        var docCount = aggregations.findValue("docCount");
+        Assertions.assertNotNull(docCount);
+        Assertions.assertEquals(2, docCount.asInt());
+
+        var snakeCase = aggregations.findValue("doc_cound_error_upper_bound");
+        assertThat(snakeCase, is(nullValue()));
+    }
+
```
sondrev, line 64: As I understand it your code only renames a few hardcoded values, and you test that theese are renamed here, but can you also add a test that makes sure we dont rename anything we are not supposed to?

search-commons/src/test/java/no/unit/nva/search/models/SearchResponseDtoTest.java
```diff
@@ -38,6 +49,24 @@ void jsonSerializationShouldKeepDeprecatedFieldsUntilFrontendHasMigrated() throw
...
+    private SearchResponse getSearchResponse() throws IOException {
+        String jsonResponse = stringFromResources(Path.of(SAMPLE_OPENSEARCH_RESPONSE_JSON));
+        return SearchResponseUtil.getSearchResponseFromJson(jsonResponse);
+    }
+
 
     private SearchResponseDto randomResponse() {
         return SearchResponseDto.builder()
```
sondrev, line 72: There should also be a test at the lambda layer that makes sure that the return value has the fields we have "promised" to return to the frontend

search-commons/src/main/java/no/unit/nva/search/models/SearchResponseDto.java
```diff
@@ -184,17 +171,65 @@ public boolean equals(Object o) {
...
+    private static ObjectNode aggregationFieldsToCamelCase(JsonNode aggregations) {
+        var json = (ObjectNode) aggregations;
+
+        for (JsonNode aggregationNode : json) {
+            var aggregation = (ObjectNode) aggregationNode;
+            AGGREGATION_FIELD_TO_CHANGE.forEach((oldname, newName) -> renameField(aggregation, oldname, newName));
+
+            for (JsonNode bucketNode : aggregation.get(BUCKETS_FIELD_NAME)) {
+                var bucket = (ObjectNode) bucketNode;
+                BUCKET_FIELD_TO_CHANGE.forEach((oldName, newName) -> renameField(bucket, oldName, newName));
```
sondrev, line 223: What happens if the field does not exist? My guess is that it will be added with null data. Is this the wanted behaviour?

sondrev (changes requested):

sondrev (approved):

PR #259, opened 2023-03-06
Np 36708 aggregation for ticket type

EyobSikt:
Created  aggregation for tickets  and added ticket.type to the aggregation.


search-commons/src/main/java/no/unit/nva/search/models/SearchTicketsQuery.java
```diff
@@ -85,4 +92,8 @@ private void addViewingScope(ViewingScope viewingScope, BoolQueryBuilder queryBu
                 .queryName(SearchClient.EXCLUDED_VIEWING_SCOPES_QUERY_NAME);
         }
     }
+
+    private void addAggregations(SearchSourceBuilder sourceBuilder) {
+        aggregations.forEach(aggDTO -> sourceBuilder.aggregation(aggDTO.toAggregationBuilder()));
```
ketilaa, line 97: We try to avoid abbreviations and use of uppercase in variable names. `aggDTO` -> `aggregation` so that is matches name of source (`aggregations`)?

search-commons/src/test/java/no/unit/nva/search/OpensearchTest.java
```diff
@@ -114,7 +114,7 @@ void shouldReturnZeroHitsOnEmptyViewingScope() throws Exception {
         indexingClient.addDocumentToIndex(getIndexDocument(indexName, Set.of()));
 
         Thread.sleep(DELAY_AFTER_INDEXING);
-        SearchTicketsQuery searchTicketsQuery = new SearchTicketsQuery(PAGE_SIZE, PAGE_NO);
+        SearchTicketsQuery searchTicketsQuery = new SearchTicketsQuery(PAGE_SIZE, PAGE_NO, null);
```
ketilaa, line 117: Instantiating an object passing null value in constructor is not recommended. In this case, Collections.emptyList() to indicate no aggregations seems like a better approach to me.

search-commons/src/test/java/no/unit/nva/search/OpensearchTest.java
```diff
@@ -432,4 +432,63 @@ private IndexDocument crateSampleIndexDocument(String indexName, String jsonFile
 
         return new IndexDocument(eventConsumptionAttributes, jsonNode);
     }
-}
\ No newline at end of file
+
+    @Test
+    void shouldNotReturnTicketsAggregationsWhenNotRequested() throws ApiGatewayException, IOException,
```
ketilaa, line 437: It seems like most tests in this class create an index and delete it afterwards. This is prone to error as any exception thrown is the test leads to an index that is never deleted. Consider refactoring index creation to be done in methods annotated with @BeforeAll and @AfterAll. Might not be a big issue in this case at the test container is deleted after the tests are run anyway, but it makes the test code more readable with less code.

ketilaa (changes requested):
Consider having a look at some of my comments.

search-commons/src/main/java/no/unit/nva/search/models/SearchTicketsQuery.java
EyobSikt: Done.

search-commons/src/test/java/no/unit/nva/search/OpensearchTest.java
EyobSikt: Done.

search-commons/src/test/java/no/unit/nva/search/OpensearchTest.java
EyobSikt: Done.

search-commons/src/test/java/no/unit/nva/search/OpensearchTest.java
```diff
@@ -246,22 +241,21 @@ void shouldVerifySearchNotReturningHitsWithDraftPublicationRequestInSearchRespon
...
         indexingClient.addDocumentToIndex(
             crateSampleIndexDocument(indexName, "sample_publishing_request_of_draft_publication.json"));
         indexingClient.addDocumentToIndex(
             crateSampleIndexDocument(indexName, "sample_publishing_request_of_published_publication.json"));
         Thread.sleep(DELAY_AFTER_INDEXING);
+
         var viewingScope = ViewingScope.create(ORGANIZATION_ID_URI_HARDCODED_IN_SAMPLE_FILES);
-        SearchTicketsQuery searchTicketsQuery = new SearchTicketsQuery(PAGE_SIZE, PAGE_NO);
+        var aggregations = ApplicationConstants.TICKETS_AGGREGATIONS;
+        SearchTicketsQuery searchTicketsQuery = new SearchTicketsQuery(PAGE_SIZE, PAGE_NO, aggregations);
```
sondrev, line 258: You should test the aggregation in a new test, rather then modifying a existing one

sondrev (changes requested):
I would like to see a test in SearchTicketsHandlerTest to check that the returned response from the lambda includes the new aggregation field in its response.

search-commons/src/test/java/no/unit/nva/search/OpensearchTest.java
```diff
@@ -86,10 +89,16 @@ static void setUp() {
     }
 
     @AfterAll
-    static void afterAll() {
+    static void afterAll() throws Exception {
+        indexingClient.deleteIndex(indexName);
```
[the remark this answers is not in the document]
EyobSikt: Since there was a method which does not need to add documents to an index (and hence getting error), I have to create nested class and add all other methods in the nested class. This way we add  @beforeEach and  @afterEach. Test execution becomes also more readable.

search-commons/src/test/java/no/unit/nva/search/OpensearchTest.java
EyobSikt: Done.  Addtional:- Tested aggregation in SearchTicketsHandlerTest.

ketilaa (approved):

sondrev (approved):

PR #260, opened 2023-03-09
Move aggregations to inspect root level

brinxmat:


search-commons/src/test/resources/sample_publication_with_affiliations.json
```diff
@@ -176,5 +170,10 @@
     "owner": "kirTest@unit.no",
     "ownerAffiliation": "https://www.example.org/Bergen"
   },
-  "status": "PUBLISHED"
+  "status": "PUBLISHED",
+  "topLevelOrganization": [{
```
sondrev, line 174: Should we rename this field to topLevelOrganizations since it now contains multiple values?

sondrev (approved):

PR #262, opened 2023-03-23
Add provision for setting mappings

brinxmat:
- Adds mappings to index creation


sondrev (changes requested):
I would like to see some tests for this. Such as a OpenSearchTest (docker) that makes sure that created indexes with null mapping are actually dynamics and not just with empty mapping.

ketilaa (approved):

ketilaa:
Approving, but is seems like the null mappings may apply some changes to the behaviour in the open search java client. I am assuming this will be detected and covered by existing tests.

PR #273, opened 2023-04-28
np-44605-swagger-200-ok

StigNorland:


docs/openapi.yaml
```diff
@@ -60,7 +60,7 @@ paths:
           content:
             application/json:
               schema:
-                type: object
+                $ref: "#/components/schemas/SearchResponse"
```
sondrev, line 63: TicketSearchResponse

docs/openapi.yaml
```diff
@@ -108,7 +108,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/Publication"
+                $ref: "#/components/schemas/SearchResponse"
```
sondrev, line 111: ResourceSearchResponse

docs/openapi.yaml
```diff
@@ -278,7 +297,53 @@ components:
...
+          $ref: '#/components/schemas/Id'
+        processingTime:
+          type: integer
+          description: time consumed on server in ms.
+        size:
+          type: integer
+          description: Number of results returned from query
+        aggregations:
+          type: object
+          description: The requested path associated with the error
```
sondrev, line 323: this seems to be copy/pastes from somewhere else

sondrev (changes requested):

PR #275, opened 2023-05-05
Np 44666 unread

hackon:


search-resources-api/src/main/java/no/unit/nva/search/SearchTicketsHandler.java
```diff
@@ -18,156 +30,162 @@
...
+    final var owner = requestInfo.getUserName();
+    logger.info("OwnerScope: {}", owner);
+    var unread = new FilterAggregationBuilder(
+        "unread",
+        notInFilter(requestInfo, owner, "owner", "viewedBy"));
+    var aggregations = List.of(
+        TYPE_TERMS_AGGREGATION, STATUS_TERMS_AGGREGATION, unread);
+    return searchClient.searchOwnerTickets(toQueryTickets(requestInfo, aggregations), owner, indexName);
+  }
+
```
EyobSikt, line 89: Looks good. But we may need to test this in lambda handler?

hackon: It's a tricky one. The other 2 are "agnostic" in the sense that if the response contains a certain field, create a bucket aggregation on it. 
This filter bucket need the same/similar query to the main query.
Ideally this would be covered with the tests running with TestContainer in the commons module, but since they requires a knowledge of the query and the values from the request it fits here.

EyobSikt (approved):

brinxmat (commented):
Weird spacing change…

PR #279, opened 2023-05-10
init handler for importCandidates

truhacevkir:


truhacevkir:
> Do we need a new handler for this? Can't the old handler be used? @brinxmat what do you think? On one hand we keep this decoupled from everything related to publication-search indexes, but we are making code-duplication.

Fikk

anetteOlli (approved):

PR #283, opened 2023-05-22
index import candidates handler created

truhacevkir:
Copy if IndexResourceHandler which uses import-candidates index

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -408,7 +408,7 @@ Resources:
                 - Lambda Function Invocation Result - Success
               detail:
                 responsePayload:
-                  topic: [ "ImportCandidates.ExpandedDataEntry.Update" ]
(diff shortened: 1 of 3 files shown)


template.yaml
```diff
@@ -382,6 +382,34 @@ Resources:
...
+          Type: EventBridgeRule
+          Properties:
+            EventBusName:
+              Ref: EventBusName
+            Pattern:
+              detail-type:
+                - Lambda Function Invocation Result - Success
+              detail:
+                responsePayload:
+                  topic: [ "ImportCandidates.ExpandedDataEntry.Update" ]
```
anetteOlli, line 411: Burde ikke dette være ImportCandidates.ExpandedDataEntry.Persisted?

anetteOlli (changes requested):

anetteOlli (approved):

PR #292, opened 2023-06-22
Add settings to indices

brinxmat:
- Adds a settiings config for the resources index
- This hopefully means that the errors on "complex" queries disappear without interfering with other functionality


brinxmat:
Does not seem relevant since it is a year old

PR #295, opened 2023-06-27
Qa pr using beans

brinxmat:
- I noted that the solution you present works fine, but I also note that we need to use a uniform, extensible mechanism to produce this kind of data
- As such I offer the following example
- Note that I have also reworked the testing somewhat, which I'm aware isn't great, but it tests the in/output more clearly
- Sorry about the size and scope of this, I don't find the testing in this project to be well-thought-out and I struggle to make things work in a way that I like.


EyobSikt:
@brinxmat Looks clean code and refactored very nicely.

EyobSikt (commented):
Looks clean code and refactored very nicely.

PR #305, opened 2023-08-24
Np 45165 filter for import candidate index

truhacevkir:


search-commons/src/test/java/no/unit/nva/search/OpensearchTest.java
```diff
@@ -170,6 +171,52 @@ private IndexDocument crateSampleIndexDocument(String indexName, String jsonFile
         return new IndexDocument(eventConsumptionAttributes, jsonNode);
     }
 
+    private void addDocumentsToIndex(String... files) throws InterruptedException {
+        Arrays.asList(files)
```
StigNorland, line 175: Stream.of(files) is marginal faster, in most cases... and you don't need a list/array

StigNorland (dismissed):
Didn't find anything that needs change.
I don't know if Stream.of() is better in this usecase, but I prefere Streams over Arrays

PR #312, opened 2023-09-14
https://unit.atlassian.net/browse/NP-45260

StigNorland:
swagger doc updated with param description


docs/openapi.yaml
```diff
@@ -78,157 +78,163 @@ paths:
       parameters:
       - in: query
         name: category
-        description: "Term to search for in resource."
+        description: "Result category code, category codes are defined in the Cristin API documentation."
```
sondrev, line 81: We shouldent redirect to a documentation that will be taken down when NVA replaced Cristin

docs/openapi.yaml
```diff
@@ -78,157 +78,163 @@ paths:
...
         name: created_since
-        description: "Term to search for in resource."
+        description: "Results created since and inclusive the given date, (yyyy-mm-dd), e.g: 2005-03-17"
         required: false
         schema:
           type: string
       - in: query
         name: doi
-        description: "Term to search for in resource."
+        description: "DOI name (e.g. doi=10.1000/123456)"
```
sondrev, line 105: Is there a example field we can use instead of putting it in the description?

docs/openapi.yaml
```diff
@@ -78,157 +78,163 @@ paths:
...
         required: false
         schema:
           type: string
       - in: query
         name: fields
-        description: "Term to search for in resource."
+        description: "'fields=all' gives a list of result objects with all available fields. If this parameter is omitted, a list of result summaries with fewer fields will be returned."
```
sondrev, line 111: what other values do we support? We use schema type 'string' underneath but here it seems like a enum or boolean? We should document all posibilities if its a enum

docs/openapi.yaml
```diff
@@ -78,157 +78,163 @@ paths:
...
         name: issn
-        description: "Term to search for in resource."
+        description: "The issn of the result"
         required: false
         schema:
           type: string
       - in: query
         name: lang
-        description: "Term to search for in resource."
+        description: "Two letter language code. Using the 'lang' parameter, this is no longer in use, and are ignored"
```
sondrev, line 147: if its no longer in use why not just remove it?

docs/openapi.yaml
```diff
@@ -78,157 +78,163 @@ paths:
...
         required: false
         schema:
           type: string
       - in: query
         name: order
-        description: "Term to search for in resource."
+        description: "SortOrder is either 'asc' or 'desc'. See Search and sort Default sort order is on 'cristin_result_id' in ascending order."
         required: false
         schema:
           type: string
```
sondrev, line 168: its not a string if theres only 2 valid values

docs/openapi.yaml
```diff
@@ -78,157 +78,163 @@ paths:
...
       - in: query
         name: year_reported
-        description: "Term to search for in resource."
+        description: "The year a result was reported"
         required: false
         schema:
           type: string
```
sondrev, line 240: should this be number?

sondrev (changes requested):

ketilaa:
@StigNorland Ping. Have a look at the feedback so you can fix, merge and finish the work.

search-commons/src/main/java/no/unit/nva/search2/ResourceAwsClient.java
```diff
@@ -102,7 +102,7 @@ private static String[] extractFields(String field) {
     private QueryBuilderSourceWrapper populateSearchRequest(QueryBuilderWrapper queryBuilderWrapper) {
         var builder = new SearchSourceBuilder().query(queryBuilderWrapper.builder());
         var query = queryBuilderWrapper.query();
-        var searchAfter = query.getValue(SEARCH_AFTER).toString();
+        var searchAfter = query.removeValue(SEARCH_AFTER);
```
StigNorland, line 105: search_after needs to be removed from normal URI and handeled on its own.

PR #315, opened 2023-09-19
https://unit.atlassian.net/browse/NP-45298

StigNorland:


search-resources-api/src/main/java/no/unit/nva/search2/ResourcePagedSearchHandler.java
```diff
@@ -35,7 +35,7 @@ protected PagedSearchResourceDto processInput(Void input, RequestInfo requestInf
         return
             ResourceQuery.Builder.queryBuilder()
                 .fromRequestInfo(requestInfo)
```
StigNorland, line 37: The actual bugfix

search-commons/src/main/java/no/unit/nva/search2/ResourceQuery.java
```diff
@@ -177,13 +175,19 @@ protected void setValue(String key, String value) {
...
-                var page = Integer.parseInt(query.getValue(PAGE));
-                var perPage = Integer.parseInt(query.getValue(PER_PAGE));
-                query.setQueryValue(OFFSET, String.valueOf(page * perPage));
+            if (nonNull(query.getValue(PAGE))) {
+                if (isNull(query.getValue(OFFSET))) {
+                    var page = Integer.parseInt(query.getValue(PAGE));
+                    var perPage = Integer.parseInt(query.getValue(PER_PAGE));
+                    query.setQueryValue(OFFSET, String.valueOf(page * perPage));
+                }
                 query.removeValue(PAGE);
```
StigNorland, line 191: bugfix-ish, PAGE must not get passed to SWS gateway and opensearch

search-commons/src/main/java/no/unit/nva/search2/ResourceQuery.java
```diff
@@ -45,14 +45,13 @@ private ResourceQuery() {
...
-        var requestParameter = toGateWayRequestParameter();
-        var id = createUriOffsetRef(url, requestParameter, offset);
-
-        var hasMoreResults = offset < response.getTotalSize();
+        final var offset = getQueryOffset();
+        final var url = gatewayUri.toString().split("\\?")[0];
+        final var requestParameter = toGateWayRequestParameter();
+        final var id = createUriOffsetRef(url, requestParameter, offset);
+        final var hasMoreResults = offset < response.getTotalSize();
         var nextResults
```
StigNorland, line 55: I tried to extract to functions, it didn't not improve readability, without doing unnecessary calls

search-commons/src/main/java/no/unit/nva/search2/ResourceQuery.java
truhacevkir: All the tests run without nested if statement:

```
@Override
        protected void applyRulesAfterValidation() {
            // convert page to offset if offset is not set
            if (nonNull(query.getValue(PAGE)) && isNull(query.getValue(OFFSET))) {
                var page = Integer.parseInt(query.getValue(PAGE));
                var perPage = Integer.parseInt(query.getValue(PER_PAGE));
                query.setQueryValue(OFFSET, String.valueOf(page * perPage));
                query.removeValue(PAGE);
            }
            // TODO check if field is set and has value 'all' then populate with all fields
        }
```

Should it be nested if statement? If yes, it would be nice to have a test for it, in case someone decides to refactor it later.

StigNorland:
```
URI.create("https://example.com/?category=hello+world&user=12%203&page=2"),
URI.create("https://example.com/?category=hello+world&user=12%203&offset=30"),
URI.create("https://example.com/?category=hello+world&user=12%203&from=30&results=10"),
```
1 test with page
1 test with offset
1 test with from

truhacevkir: But there is no failing tests when we combine this two if statements, which means that they do not need to be nested?

StigNorland: added these -> 
```
 void uriParamsToResourceParams()...
 
        assertNotNull(resourceParameters.getValue(OFFSET));
        assertNull(resourceParameters.getValue(PAGE));
```

truhacevkir: But should this if statement be nested? Like
```
if() {
   if() {
  }
}
```

truhacevkir (approved):

PR #316, opened 2023-09-19
https://unit.atlassian.net/browse/NP-45300

StigNorland:
category  = entityDescription.reference.publicationInstance
category -> entityDescription.reference.publicationInstance.type


search-commons/src/main/java/no/unit/nva/search2/ResourceParameterKey.java
```diff
@@ -57,7 +57,7 @@ public enum ResourceParameterKey implements ParameterKey {
     USER(STRING, "user", "resourceOwner.owner"),
     YEAR_REPORTED(NUMBER, "year_reported", "entityDescription.publicationDate.year"),
     // Query parameters passed to SWS/Opensearch
-    SEARCH_ALL(CUSTOM, NONE, "query", "q"),
+    SEARCH_ALL(CUSTOM, NONE, "query", ""),
```
StigNorland, line 60: This will be removed altogether, when frontend has transformed from old to new q params.

PR #319, opened 2023-09-27
fixed documentation and naming

StigNorland:


docs/openapi.yaml
```diff
@@ -70,7 +70,273 @@ paths:
               schema:
                 $ref: "#/components/schemas/TicketSearchResponse"
 
+  /resources:
```
torarnet, line 73: was this not reachable before?

PR #332, opened 2023-10-13
Batch index handler

truhacevkir:
- The handler reads keys from persisted resources bucket
- Keys are persisted in second bucket in batches of 10

### DIFF
--- a/batch-index/src/test/java/no/unit/nva/search/GenerateKeyBatchesHandlerTest.java
+++ b/batch-index/src/test/java/no/unit/nva/search/GenerateKeyBatchesHandlerTest.java
@@ -7,6 +7,7 @@ import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.is;
 import java.io.ByteArrayOutputStream;
 import java.net.URI;
+import java.nio.file.Path;
 import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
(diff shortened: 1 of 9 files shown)


template.yaml
```diff
@@ -629,3 +648,14 @@ Resources:
...
+  KeyBatchesBucket:
+    Type: AWS::S3::Bucket
+    Properties:
+      AccessControl: Private
+      BucketName: !Sub "${KeyBatchesBucketName}-${AWS::AccountId}"
+      LifecycleConfiguration:
+        Rules:
+          - Id: DeleteContentAfter3Days
+            Status: Enabled
+            ExpirationInDays: 3
```
anetteOlli, line 661: NB! make sure this is not too short span. We have had issues with this in cristin-import.

template.yaml
```diff
@@ -582,13 +594,20 @@ Resources:
           BATCH_INDEX_EVENT_BUS_NAME: !GetAtt BatchIndexEventBus.Name
           PERSISTED_RESOURCES_PATH: !Sub "s3://${ResourcesBucket}/resources"
 
-  BatchIndexEventBus:
-    Type: AWS::Events::EventBus
+  GenerateKeyBatchesHandler:
```
anetteOlli, line 597: Can this name indicate that it is for the search-index?

template.yaml
```diff
@@ -463,6 +467,14 @@ Resources:
                 responsePayload:
                   topic: [ "PublicationService.ExpandedEntry.Persisted" ]
 
+  BatchIndexEventBus:
```
anetteOlli, line 470: Note that max age for events on aws events bus is 6 hours

anetteOlli (dismissed):

anetteOlli (approved):

batch-index/src/test/java/no/unit/nva/search/GenerateKeyBatchesHandlerTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+        var actual = getPersistedFileFromOutputBucket();
+        var expectedNumberOfFiles = getExpectedNumberOfFiles(numberOfItemsInBucket);
+
+        assertThat(actual.size(), is(equalTo(expectedNumberOfFiles)));
+        assertThat(actual.stream().collect(Collectors.joining(System.lineSeparator())), is(equalTo(expected)));
+    }
+
+    private static int getExpectedNumberOfFiles(int numberOfItemsInBucket) {
+        return (int) Math.ceil(numberOfItemsInBucket < 11 ? 1 : (double) numberOfItemsInBucket / 10);
```
sondrev, line 59: 10 should be a constant

This method was hard to follow. Maybe split the edgecase of < 11 outside of the Math.ceil?

PR #339, opened 2023-10-19
https://unit.atlassian.net/browse/NP-45237

StigNorland:
mediatype cause trouble fix

### DIFF
--- a/search-commons/src/main/java/no/unit/nva/search2/ResourceAwsQuery.java
+++ b/search-commons/src/main/java/no/unit/nva/search2/ResourceAwsQuery.java
@@ -49,7 +49,7 @@ public final class ResourceAwsQuery extends OpenSearchQuery<ResourceParameterKey
 
     public String doSearch(ResourceAwsClient queryClient) {
         return
-            this.getMediaType().equals(MediaType.CSV_UTF_8)
+            this.getMediaType().is(MediaType.CSV_UTF_8)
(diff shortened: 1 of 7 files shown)


search-commons/src/main/java/no/unit/nva/search2/ResourceAwsQuery.java
```diff
@@ -49,7 +49,7 @@ static Builder builder() {
 
     public String doSearch(ResourceAwsClient queryClient) {
         return
-            this.getMediaType().equals(MediaType.CSV_UTF_8)
+            this.getMediaType().is(MediaType.CSV_UTF_8)
```
torarnet, line 52: nullsafe?

StigNorland: Nei, men getMediaType er alltid fyllt inn

StigNorland: Jeg snur uttrykket for ordens skyld

torarnet (approved):
Maybe write a unit test for different accept headers, otherwise OK

StigNorland:
> Maybe write a unit test for different accept headers, otherwise OK

It is tested in the Handler which is in another subproject  `shouldReturnTextCsvWithGivenAcceptHeader`

PR #345, opened 2023-10-23
https://unit.atlassian.net/browse/NP-45471

StigNorland:
+ funding_source / funding fix

### DIFF
--- a/search-commons/src/main/java/no/unit/nva/search2/ResourceAwsClient.java
+++ b/search-commons/src/main/java/no/unit/nva/search2/ResourceAwsClient.java
@@ -57,7 +57,7 @@ public class ResourceAwsClient implements OpenSearchClient<OpenSearchSwsResponse
     private final HttpClient httpClient;
     private final BodyHandler<String> bodyHandler;
 
-    private final Integer SINGLE_FIELD = 1;
+    private static final Integer SINGLE_FIELD = 1;
 
     public ResourceAwsClient(CachedJwtProvider cachedJwtProvider, HttpClient client) {
         super();
@@ -87,13 +87,13 @@ public class ResourceAwsClient implements OpenSearchClient<OpenSearchSwsResponse
 
     private Stream<QueryBuilderWrapper> createQueryBuilderStream(ResourceAwsQuery query) {
         AbstractQueryBuilder<?> queryBuilder;
-            if (query.isPresent(SEARCH_ALL)) {
-                queryBuilder = multiMatchQuery(query);
-            } else if (query.noLucineParameter()) {
-                queryBuilder = QueryBuilders.matchAllQuery();
-            } else {
-                queryBuilder = boolQuery(query);
-            }
+        if (query.isPresent(SEARCH_ALL)) {
+            queryBuilder = multiMatchQuery(query);
+        } else if (query.noLucineParameter()) {
+            queryBuilder = QueryBuilders.matchAllQuery();
+        } else {
+            queryBuilder = boolQuery(query);
+        }
 
         return Stream.of(new QueryBuilderWrapper(queryBuilder, query));
     }
@@ -116,7 +116,6 @@ public class ResourceAwsClient implements OpenSearchClient<OpenSearchSwsResponse
         builder.from(query.getValue(FROM).as());
         getSortStream(query).forEach(orderTuple -> builder.sort(orderTuple.v1(), orderTuple.v2()));
 
-
         return new QueryBuilderSourceWrapper(builder, query.openSearchSwsUri());
     }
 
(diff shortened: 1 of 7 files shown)


search-commons/src/main/java/no/unit/nva/search2/constant/Patterns.java
```diff
@@ -17,8 +17,12 @@ public class Patterns {
      */
     public static final String PATTERN_IS_DATE_STRING = "\\b\\d{4}(?:-\\d{2}-\\d{2})?\\b";
 
+
+    public static final String PATTERN_IS_ADD_SLASH = "\\\\$1";
+    public static final String PATTERN_IS_IGNORE_CASE = "(?i)";
+    public static final String PATTERN_IS_NONE_OR_ONE = ".?";
     public static final String PATTERN_IS_NON_EMPTY = ".+";
     public static final String PATTERN_IS_NUMBER = "[0-9]\\d*";
-    public static final String PATTERN_IS_NONE_OR_ONE = ".?";
```
StigNorland, line 22: not deleted, just sorted

search-commons/src/main/java/no/unit/nva/search2/ResourceAwsQuery.java
```diff
@@ -212,8 +208,13 @@ private String expandFields(String value) {
             return ALL.equals(value) || isNull(value)
                 ? ALL
                 : Arrays.stream(value.split(COMMA))
-                .filter(Builder::notInvalid)
```
StigNorland, line 215: removed static, bad for performance?

torarnet: I wonder about static myself. Does every caller share the static method and have to wait in turn to use it, or is it better because it is loaded only once in memory and every caller can use it in parallell? We have discussed this and we are not really sure

search-commons/src/main/java/no/unit/nva/search2/model/ParameterKey.java
```diff
@@ -26,7 +28,7 @@ static Predicate<ParameterKey> equalTo(String name) {
     }
 
     static String escapeSearchString(String value) {
-        return value.replaceAll("([-+&|!(){}\\[\\]^\"~*?:\\\\/])", "\\\\$1");
+        return value.replaceAll(PATTERN_IS_SPECIAL_CHARACTERS, PATTERN_IS_ADD_SLASH);
```
torarnet, line 31: Is this the magic strings solving the task?

search-commons/src/main/java/no/unit/nva/search2/model/ParameterKey.java
```diff
@@ -39,11 +41,11 @@ enum ParamKind {
 
     enum Operator {
         NONE("%s%s"),
-        EQUALS("%s:%s"),
-        GREATER_THAN("%s:>%s"),
-        GREATER_THAN_OR_EQUAL_TO("%s:>=%s"),
-        LESS_THAN("%s:<%s"),
-        LESS_THAN_OR_EQUAL_TO("%s:<=%s"),
+        EQUALS("(%s:%s)"),
```
torarnet, line 44: why these ( ) ?

search-commons/src/main/java/no/unit/nva/search2/ResourceAwsClient.java
```diff
@@ -161,7 +140,54 @@ private OpenSearchSwsResponse handleResponse(HttpResponse<String> response) {
...
+    private BoolQueryBuilder boolQuery(ResourceAwsQuery query) {
+        var bq = QueryBuilders.boolQuery();
+        query.getLuceneParameters()
+            .forEach((key, value) -> {
+                var swsKey = key.swsKey().toArray(String[]::new);
+                if (swsKey.length > SINGLE_FIELD) {
+                    bq.must(QueryBuilders.multiMatchQuery(value, swsKey));
+                } else {
+                    if (key.operator() != ParameterKey.Operator.EQUALS) {
+                        bq.must(rangeQuery(key.operator(), swsKey[0], value));
```
torarnet, line 162: Can you explain these new changes to boolQuery??

torarnet: A lot of logic here is tightly coupled to the search engine, and is hard for me to understand when i dont know how search works. 

Maybe consider adding some javadoc for readability... Not a must have, but makes it easier to understand

search-commons/src/main/java/no/unit/nva/search2/model/ParameterKey.java
StigNorland: This is i list of charcters that requiers an added escape '\'  inorder to work in opensearch.

( [ &! ] ) -> search for '&' or '!', when encountered add to (a) find group, and replace with '\+(fond group)'
anyone of these characters should be replaced by  '/' + (character)

If you have better names or a solutions that doesn't requires 50 lines of code, and doesn't have to iterate text for each char,   I am listening

search-commons/src/main/java/no/unit/nva/search2/model/ParameterKey.java
StigNorland: I was suprised to see this in my code here. Thought it was in a earlier attempt to use lucene, and that it didn't go to production.
This is not in use at the moment, can change it back.

torarnet (approved):

PR #355, opened 2023-11-01
Changes to search parameters

elkebab:
Some changes to the search parameters of `/resources2`.

Open API spec is aligned to Cristin instead of NVA atm, which also should be addressed at some point in the future.

These changes are just meant to indicate what should be different, and someone with more knowledge about the backend/search setup must verify that everything works as expected.


search-commons/src/main/java/no/unit/nva/search2/model/ResourceParameterKey.java
```diff
@@ -53,15 +53,13 @@ public enum ResourceParameterKey implements ParameterKey {
...
-    PUBLICATION_ID(STRING, "entityDescription.reference.publicationInstance.id"),
-    PUBLICATION_TYPE(STRING, "entityDescription.reference.publicationInstance.type"),
+    PROJECT(STRING, "projects.id"),
     PUBLISHED_BEFORE(DATE_STRING, LESS_THAN, "publishedDate"),
     PUBLISHED_SINCE(DATE_STRING, GREATER_THAN_OR_EQUAL_TO, "publishedDate"),
     TITLE(STRING, "entityDescription.mainTitle", 2F),
     UNIT(STRING, "entityDescription.contributors.affiliation.id"),
     USER(STRING, "resourceOwner.owner"),
-    YEAR_REPORTED(NUMBER, "entityDescription.publicationDate.year"),
+    PUBLICATION_YEAR(NUMBER, "entityDescription.publicationDate.year"),
```
StigNorland, line 62: `PUBLICATION_YEAR(NUMBER, EQUALS, "entityDescription.publicationDate.year","(?i)year.?reported|publication.?year", null, null),`
we need to support 'year_reported' as well

StigNorland (changes requested):

search-commons/src/main/java/no/unit/nva/search2/model/ResourceParameterKey.java
StigNorland: (?i)  = regex ignore case
 year.?reported -> year + any or no character + reported = yearReported /  year_reported / year reported / etc
 | -> OR
 publication.?year -> publication + any or no character + year = publicationYear/  publication_year/ publication year/ etc

StigNorland (approved):

PR #362, opened 2023-11-10
type aggregation for import-candidate index

truhacevkir:


indexing-handlers/src/main/resources/import_candidates_mappings.json
```diff
@@ -1,5 +1,26 @@
...
+        "reference": {
+          "type": "nested",
+          "include_in_parent": true,
+          "properties": {
+            "publicationInstance": {
+              "type": "nested",
+              "include_in_parent": true,
+              "properties": {
+                "type": {
+                  "type": "keyword"
```
sondrev, line 16: Do we need this searchable in the future? We have had issues with keywords in search-api where we last year changed text->keyword in mapping to make facets work, but when we later wanted to improve searching publications we found out that Keywords cant be partialy searched (Needs to be 100% excact)

To fix this you could instead do:

`                  
                "type": {
                  "type": "text",
                  "fields" : {
                    "keyword" : {
                      "type" : "keyword"
                    }
                  }
                }`
Which makes a subfield which will only be used for the aggregating leaving the main-field for search. To access this you use it like: entityDescription.reference.publicationINstance.type.keyword
You wont need to change the data

search-commons/src/test/resources/test_import_candidates_mappings.json
```diff
@@ -1,5 +1,26 @@
...
+              "properties": {
+                "type": {
+                  "type": "keyword"
+                }
+              }
+            }
+          }
+        }
+      }
+    },
```
sondrev, line 23: remember to also change here, if you decide to change the other mapping file

sondrev (approved):

PR #365, opened 2023-11-17
Np 45670 reindex specified index with bulk

truhacevkir:
Works as before for `resources` index.
Need to send `location` as event param to reindex `import-candidates` index
Validation is turned on only for `resources` index, cause `import-candidates` have other json structure, and this is something we can discuss/implement later. 

Have also added some documentation on how to use it, give feedback if something could be explained better.

https://unit.atlassian.net/wiki/spaces/NVAP/pages/2863792151/Indexing


StigNorland (approved):

batch-index/src/test/java/no/unit/nva/search/keybatch/GenerateKeyBatchesHandlerTest.java
```diff
@@ -59,9 +59,22 @@ void setUp() {
 
     @Test
     void shouldPersistS3KeysToBatchBucket() throws JsonProcessingException {
-        final var allFiles = putObjectsInInputBucket(SINGLE_BATCH_FILE_SIZE);
+        final var allFiles = putObjectsInInputBucket(SINGLE_BATCH_FILE_SIZE, DEFAULT_LOCATION);
```
StigNorland, line 62: it used to be called s3location, if this is aws location, why set it in code?

StigNorland (changes requested):

batch-index/src/test/java/no/unit/nva/search/keybatch/GenerateKeyBatchesHandlerTest.java
truhacevkir: This is not a s3 bucket, but s3 path, which can have two values - `resources` and `import-candidates`.
In this way we can use this lambda to generate batches for two indexes, where `resources` is default index.
It is optional variable that can be set from input event :)

brinxmat (approved):

PR #379, opened 2023-12-18
https://unit.atlassian.net/browse/NP-45875

StigNorland:
Mappings updated to use KEYWORD in all needed queries


truhacevkir (commented):
Assuming that existing tests are testing enough, do we have a test which tests that we can perform partial search?

for example shouldReturnHitsWhenSeachingBySurname?

truhacevkir (approved):

PR #380, opened 2023-12-18
https://unit.atlassian.net/browse/NP-45875

StigNorland:
frontend needs to freetext search in contributorName, changed to TEXT and added one test case.


search-commons/src/main/java/no/unit/nva/search2/common/OpenSearchClient.java
```diff
@@ -47,7 +47,7 @@ protected HttpResponse<String> fetch(HttpRequest httpRequest) {
     }
 
     protected HttpRequest createRequest(QueryContentWrapper qbs) {
-        logger.info(qbs.source().query().toString());
+        logger.debug(qbs.source().query().toString());
```
StigNorland, line 50: this is no longer needed on the server, but is nice to have in devlopment

elkebab (approved):

PR #390, opened 2024-01-10
filter curator search based on accessrights

sondrev:


StigNorland (approved):
You should consider using filter instead, as this will always be executed (filter is cached)

PR #400, opened 2024-01-25
NP-46073 Rename publishStatus to status, add openapi documentations f…

joachimjorgensen:
…or params on new endpoint


StigNorland (approved):
Sondre found this...
https://stackoverflow.com/questions/27005105/swagger-openapi-use-ref-to-pass-a-reusable-defined-parameter

StigNorland (changes requested):
https://stackoverflow.com/questions/27005105/swagger-openapi-use-ref-to-pass-a-reusable-defined-parameter

StigNorland (approved):
Nice

PR #404, opened 2024-01-30
remove old ref in openapi.yaml

StigNorland:


StigNorland:
changes done in pair programing with Rurik

PR #413, opened 2024-02-05
Hotfix/fix promoted

StigNorland:
Promoted had several issues.

1. URI in usersettings, not the same as publish.id
2. to remedy, extract uuid.
3. UUID.fromString(uuidFromPromoted) UUID too large(?)
4. ID is with URI, indentifiers just uuid...

fixed -> 
extract uuid
search on identifier instedof id

### DIFF
--- a/search-commons/src/main/java/no/unit/nva/search2/ResourceQuery.java
+++ b/search-commons/src/main/java/no/unit/nva/search2/ResourceQuery.java
@@ -56,7 +56,6 @@ import no.unit.nva.search2.enums.ParameterKey.ValueEncoding;
 import no.unit.nva.search2.enums.PublicationStatus;
 import no.unit.nva.search2.enums.ResourceParameter;
 import nva.commons.core.JacocoGenerated;
-import nva.commons.core.paths.UriWrapper;
 import org.opensearch.index.query.BoolQueryBuilder;
 import org.opensearch.index.query.QueryBuilders;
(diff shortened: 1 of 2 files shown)


search-commons/src/main/java/no/unit/nva/search2/ResourceQuery.java
```diff
@@ -228,11 +226,12 @@ private void addPromotedPublications(UserSettingsClient userSettingsClient, Bool
...
-                    QueryBuilders
-                        .matchQuery(ID, promotedPublications.get(i))
-                        .boost(3.14F + promotedPublications.size() - i)
-                );
+                var uuid = fromUri(promotedPublications.get(i)).getLastPathElement();
+                var qb = QueryBuilders
+                    .matchQuery(IDENTIFIER_KEYWORD, uuid)
+                    .boost(3.14F + promotedPublications.size() - i);
+                logger.info(qb.toString());
+                bq.should(qb);
```
truhacevkir, line 234: UUID is probably Sortableidentifier, which have timestamp "injected" in it.

truhacevkir (approved):

search-commons/src/main/java/no/unit/nva/search2/ResourceQuery.java
StigNorland: That explains UUID too large....

PR #420, opened 2024-02-12
refactoring enums, no code change

StigNorland:


search-commons/src/main/java/no/unit/nva/search2/common/Query.java
```diff
@@ -341,7 +341,7 @@ private void logSearchKeys() {
         logger.info(
             getSearchParameterKeys()
                 .map(Enum::name)
-                .collect(Collectors.joining(", ", "{\"keys\":[", "]}"))
+                .collect(Collectors.joining("\", \"", "{\"keys\":[\"", "\"]}"))
```
StigNorland, line 344: ups. code change, added `"` around keyword to make log entry into a valid json

PR #429, opened 2024-02-16
adding nvi reporting period facet and search params

truhacevkir:


search-commons/src/main/java/no/unit/nva/search2/common/AggregationFormat.java
```diff
@@ -137,7 +138,8 @@ static final class Constants {
             CONTRIBUTOR, "/filter/entityDescription/contributor/id",
             CONTEXT_TYPE, "/filter/entityDescription/reference/publicationContext/contextType",
             FUNDING_SOURCE, "/filter/fundings/id",
-            TOP_LEVEL_ORGANIZATION, "/filter/topLevelOrganizations/id"
+            TOP_LEVEL_ORGANIZATION, "/filter/topLevelOrganizations/id",
+            SCIENTIFIC_INDEX, "/filter/scientificIndex/year"
```
truhacevkir, line 142: It was really hard to find this thing. It seems like it must include aggregation path so aggregation does not filters out from original OpenSearch response?
Can this thing be made more visible?

StigNorland: Agree! I am willing to try any suggestions

search-commons/src/main/java/no/unit/nva/search2/resource/ResourceParameter.java
```diff
@@ -90,6 +90,8 @@
 public enum ResourceParameter implements ParameterKey {
     INVALID(ParameterKind.INVALID),
     // Parameters used for filtering
+    SCIENTIFIC_REPORT_PERIOD_SINCE(CUSTOM, FieldOperator.GREATER_THAN_OR_EQUAL_TO, SCIENTIFIC_INDEX_YEAR),
+    SCIENTIFIC_REPORT_PERIOD_BEFORE(CUSTOM, FieldOperator.LESS_THAN, SCIENTIFIC_INDEX_YEAR),
```
StigNorland, line 94: You can use Number instead of CUSTOM, if not handled, CUSTOM will fall through to keyword, which works fine now, but might break in thre future.

StigNorland (approved):

search-commons/src/main/java/no/unit/nva/search2/common/AggregationFormat.java
truhacevkir: What about implementing something like this when mapping sws response to PagedSearch resposne?
Instead of:

```
.withAggregations(response.getAggregationsStructured())
                .build();
```

```
.withAggregations(response.keepRequiredAggregations(REQUIRED_AGGREGATIONS).getStructuredAggregations())
                .build();
```

It makes it visible early in mapping process that we are filtering something out.

truhacevkir: If you agree, i can try to do something similar in next pr.

search-commons/src/main/java/no/unit/nva/search2/common/constant/Words.java
```diff
@@ -32,6 +34,8 @@ public final class Words {
     public static final String EQUAL = "=";
     public static final String FROM = "from";
     public static final String FUNDING = "FUNDING";
+    public static final String SCIENTIFIC_REPORT_PERIOD_START = "SCIENTIFIC_REPORT_PERIOD_START";
+    public static final String SCIENTIFIC_REPORT_PERIOD_END = "SCIENTIFIC_REPORT_PERIOD_END";
```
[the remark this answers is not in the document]
truhacevkir: Have removed those

PR #445, opened 2024-02-29
added test source files

StigNorland:
Just to reduce the size of scaffolding pr


truhacevkir (approved):

truhacevkir:
It is pretty hard to have control over all the documents in single file when testing, which makes it pretty hard to test single feature. I would not do it that way, but it is up to you :)

PR #446, opened 2024-03-01
pre tickets update

StigNorland:
In order to use `AggregationFormat ` on more than `resources`, we need to have definisions for each entity and a way to apply them....


search-commons/src/main/java/no/unit/nva/search2/resource/ResourceQuery.java
```diff
@@ -151,6 +155,11 @@ protected boolean isPagingValue(ResourceParameter key) {
         return key.ordinal() >= FIELDS.ordinal() && key.ordinal() <= SORT_ORDER.ordinal();
     }
 
+    @Override
+    protected Map<String, String> aggregationsDef() {
```
truhacevkir, line 159: Why not to name with full name? 
`aggregationDefinitons()`
All these abbreviations does not make a lot of sense for one who does not know the context

search-commons/src/main/java/no/unit/nva/search2/common/Query.java
```diff
@@ -108,14 +111,15 @@ protected String toCsvText(SwsResponse response) {
     public PagedSearch toPagedResponse(SwsResponse response) {
         final var requestParameter = toNvaSearchApiRequestParameter();
         final var source = URI.create(getNvaSearchApiUri().toString().split(PATTERN_IS_URL_PARAM_INDICATOR)[0]);
+        final var aggregationFormatted = AggregationFormat.apply(response.aggregations(),aggregationsDef()).toString();
```
truhacevkir, line 114: Isn't AggregationFormat actually "our" format of aggregation returned after we have converted OpenSearch original aggregation to "our own" format? By adding JsonNode aggregation field to AggregationFormat and renaming AggregattionFormat to just Aggregation, we can achieve more sense:

For example:

`Aggregation.fromOpenSearchAggregation(jsonNode).toJsonString()`

I think it makes more sense now since you have extracted `AggregationFormat` from `SwsResponse.class`

Since we know the context we can change it to ResourceAggregation, and tickets will have TicketsAggregation. In that we you do not need to pass aggregationDefs to Aggregation, since Resource Aggregation should know its own definitions.

truhacevkir (approved):

search-commons/src/main/java/no/unit/nva/search2/common/Query.java
StigNorland: That should be a separate task I think, but yes.

PR #468, opened 2024-03-21
Np add recovery queue to indexing handler

truhacevkir:


indexing-handlers/src/main/java/no/unit/nva/indexing/handlers/IndexResourceHandler.java
```diff
@@ -50,21 +53,26 @@ protected Void processInputPayload(EventReference input,
...
-        return new RuntimeException(exception);
+    private Void persistRecoveryMessage(Failure<Void> failure, IndexDocument indexDocument) {
+        var documentIdentifier = indexDocument.getDocumentIdentifier();
+        var indexName = indexDocument.getIndexName();
+        RecoveryEntry.withIdentifier(documentIdentifier)
+            .withIndex(indexName)
+            .withException(failure.getException())
+            .persist(queueClient);
+        LOGGER.error(SENT_TO_RECOVERY_QUEUE_MESSAGE, indexName, documentIdentifier);
+        return null;
```
anetteOlli, line 71: Is it intentional that all exceptions are swallowed here? if the exception is completely swallowed nothing will be put on the DLQ.

truhacevkir: Message will be out on RecoveryQueue, which will be used to rerun all the entries that have failed.

ketilaa: I expect recovery to "give up" at some point, and then we need some traces to end up in a DLQ.

truhacevkir: Given what is implemented, recovery queue will never produce anything automatically. Messages from this queue will be proceeded manually by triggering "Recovery Batch Scan".
We can throw exception after delivering message to recovery queue, so defaut dlq will make multiple attempts, but in case it fails multiple times -> it will deliver multiple recovery messages for the same resource to recovery queue, which is not what we want.

anetteOlli: If that's the case then the DLQ on IndexResourceHandler should be removed from the template.

And also there is a case where this will fail: when it can't find the object in the s3 bucket just before the start of this attempt. (Correct me if I am wrong).

truhacevkir: That's right, if s3 object is not present, entry will be proceeded to dlq and not recovery queue, and thats great?
Since all the entries proceeded to the recovery queue must be identifiable (have id), so we can rerun than entry by running batch scan.

brinxmat (approved):
Approving, interested to see test of this

PR #475, opened 2024-04-03
Hotfix/queryfilter map

StigNorland:


search-commons/src/main/java/no/unit/nva/search2/common/QueryFilter.java
```diff
@@ -4,28 +4,33 @@
...
         return boolQueryBuilder;
     }
 
     public void set(QueryBuilder... filters) {
         this.filters.clear();
-        this.filters.addAll(List.of(filters));
+        var filterMap = Arrays.stream(filters)
+            .map(filter -> Map.entry(filter.queryName(), filter))
+            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+        this.filters.putAll(filterMap);
```
sondrev, line 30: perhaps something like

        this.filters.clear();
        Arrays.stream(filters)
            .forEach(this::add)

sondrev: maybe jsut .foreach even? no stream

sondrev (approved):

search-commons/src/main/java/no/unit/nva/search2/common/QueryFilter.java
StigNorland: One would think that was a natural function for an array, but in Java, **no** 🤷

PR #480, opened 2024-04-08
rollback 'flat_object'

StigNorland:


search-commons/src/main/java/no/unit/nva/search2/common/Query.java
```diff
@@ -123,7 +123,7 @@ public Stream<QueryContentWrapper> createQueryBuilderStream() {
 
         getSortStream().forEach(builder::sort);
 
-        if (getMediaType().is(JSON_UTF_8)) {
+        if (getMediaType().is(JSON_UTF_8) && isDefined(ALL)) {
```
StigNorland, line 126: ups, should have been here, but it is harmless, (it will not go through all aggregations code, if 'all' is not supplied)

sondrev (changes requested):
We are upgrading opensearch so we probably dont need this rollback

PR #484, opened 2024-04-12
access for users without access rights

StigNorland:


search-commons/src/main/java/no/unit/nva/search2/ticket/TicketQuery.java
```diff
@@ -106,12 +104,21 @@ public static TicketParameterValidator builder() {
      * @return TicketQuery (builder pattern)
      */
     public TicketQuery applyContextAndAuthorize(RequestInfo requestInfo) throws UnauthorizedException {
+        if (Objects.isNull(requestInfo.getUserName())) {
+            throw new UnauthorizedException();
+        }
+
         var organization = requestInfo.getTopLevelOrgCristinId()
             .orElse(requestInfo.getPersonAffiliation());
+        var adminAccess = getAccessRights(requestInfo);
```
joachimjorgensen, line 113: Is admin the right word? I would think admin inherently has all ticket access rights.

```suggestion
        var ticketAccessRights = getAccessRights(requestInfo);
```

search-commons/src/main/java/no/unit/nva/search2/ticket/TicketQuery.java
```diff
@@ -106,12 +104,21 @@ public static TicketParameterValidator builder() {
...
 
-        return withFilterTicketType(validateAccessRight(requestInfo))
-            .withFilterOrganization(organization)
-            .withFilterCurrentUser(requestInfo.getUserName());
+        if (adminAccess.isEmpty()) {
+            withCurrentUser(requestInfo.getUserName())
+                .applyFilterCurrentUser();
+        } else {
+            withFilterTicketType(adminAccess.toArray(TicketType[]::new));
+        }
```
joachimjorgensen, line 120: Would people with DOI-access rights see their own support tickets? A curator would maybe not have support tickets, but just as an example.

search-commons/src/main/java/no/unit/nva/search2/ticket/TicketQuery.java
StigNorland: That would be a natural thing to think, I agree, but that is wrong, Admin doesn't give you anything...

joachimjorgensen (dismissed):

joachimjorgensen (approved):

PR #490, opened 2024-04-19
Np 46320 documentation

StigNorland:


StigNorland:
added descriptions to all parameters and removed old tickets endpint from openapi.yaml

PR #492, opened 2024-04-22
Large exports

sondrev:


search-resources-api/src/main/java/no/unit/nva/search2/ExportResourceHandler.java
```diff
@@ -0,0 +1,60 @@
...
+    String processS3Input(Void input, RequestInfo requestInfo, Context context) throws BadRequestException {
+        return
+            ResourceQuery.builder()
+                .fromRequestInfo(requestInfo)
+                .withRequiredParameters(FROM, SIZE, AGGREGATION)
+                .validate()
+                .build()
+                .withRequiredStatus(PUBLISHED, PUBLISHED_METADATA)
+                .withOnlyCsvFields()
+                .doExport(opensearchClient);
```
StigNorland, line 53: Since you have a separate method for export, do you need to explicit set CsvFields?

search-commons/src/main/java/no/unit/nva/search2/resource/ResourceQuery.java
```diff
@@ -133,6 +137,17 @@ public ResourceQuery withRequiredStatus(PublicationStatus... publicationStatus)
         return this;
     }
 
+    public ResourceQuery withoutRange() {
+        this.parameters().set(FROM, "0");
```
StigNorland, line 141: Will this be ignored when using scroll?

StigNorland (approved):
Just verify that page= 0 doesn't interfere with (next) scroll, and I am happy

PR #504, opened 2024-05-02
hotfix/NP-46691-range.query

StigNorland:


search-commons/src/main/java/no/unit/nva/search2/common/builder/OpensearchQueryRange.java
```diff
@@ -36,14 +44,58 @@ protected Stream<Entry<K, QueryBuilder>> queryAsEntryStream(K key, String... val
...
+        if (key.fieldType() != ParameterKind.DATE) {
+            return retval;
+        }
+        if (retval.length() == 4) {
+            retval += "-01";
+        }
+        if (retval.length() == 7) {
+            retval += "-01";
+        }
+        return retval;
```
joachimjorgensen, line 85: Would it be necessary to validate input length as well? I.e. should `retval == "20240503"` be valid? Not sure how things work, or if the docs specify the specific format, but maybe worth a thought.

search-commons/src/main/java/no/unit/nva/search2/ticket/TicketSearchQuery.java
```diff
@@ -273,5 +274,12 @@ protected void setValue(String key, String value) {
         protected boolean isKeyValid(String keyName) {
             return TicketParameter.keyFromString(keyName) != TicketParameter.INVALID;
         }
+
+        private String expandMissing(String value) {
```
joachimjorgensen, line 278: Just a suggestion to make the method more self explanatory. Don't have to change if you don't want. 

```suggestion
        private String expandIfMissingRange(String value) {
```

search-commons/src/test/resources/datasource_urls_ticket.json
```diff
@@ -13,7 +15,7 @@
   "ID_NOT=018de5060dd0-3e2570f5-3d56-4479-b79e-4410b81626c7,018de5044416-7a3b7a74-978f-45fb-a743-c9c0b0d39e65": 18,
   "MESSAGES=en+DOI": 1,
   "MESSAGES_NOT=en+DOI": 19,
-  "MODIFIED_DATE=1872-01-01,2024": 1,
+  "MODIFIED_DATE=1872-01-01,2023": 1,
```
joachimjorgensen, line 18: Add a test for the last case in expandDateFirst/Last() methods?

```suggestion
  "MODIFIED_DATE=1872-01-01,2023": 1,
  "MODIFIED_DATE=1872-01-01,2023-05": <insert correct number here>,
```

joachimjorgensen (dismissed):

search-commons/src/main/java/no/unit/nva/search2/common/builder/OpensearchQueryRange.java
StigNorland: All parameters are validated before their use (in ParameterValidator.java)

PR #523, opened 2024-06-07
Np 46571/np 46883 accent

StigNorland:


search-commons/src/test/java/no/unit/nva/search/OpensearchTest.java
StigNorland, line 1: Time to prune old code....
Removing Opensearchtest, as new changes break these tests, and all search code is now on search2.
More pruning in the pipeline.

indexing-handlers/src/main/java/no/unit/nva/indexing/model/IndexRequest.java
```diff
@@ -23,14 +25,30 @@ public IndexRequest(String name, String jsonMappings) {
...
     }
 
+    public IndexRequest(String name, String jsonMappings, String jsonSettings) {
+        this.name = name;
+        var typeReference = new TypeReference<Map<String, Object>>() {
+        };
+        this.mappings = attempt(() -> JsonUtils.dtoObjectMapper.readValue(jsonMappings, typeReference))
+            .orElseThrow();
+        this.settings = attempt(() -> JsonUtils.dtoObjectMapper.readValue(jsonSettings, typeReference))
+            .orElseThrow();
```
brinxmat, line 38: Had to do a double-take here as I imagined I was seeing double :D Can you extract the mapping + type reference to its own method `mapToJavaType(String jsonString)`

brinxmat (changes requested):

brinxmat (approved):

PR #538, opened 2024-07-16
"simplified" code

StigNorland:
removed instance implementations from generic base classes.


search-resources-api/src/main/java/no/unit/nva/search2/SearchTicketAuthHandler.java
```diff
@@ -58,6 +60,8 @@ protected List<MediaType> listSupportedMediaTypes() {
 
     @Override
     protected void validateRequest(Void unused, RequestInfo requestInfo, Context context) throws ApiGatewayException {
-        //Do nothing
+        if (Objects.isNull(requestInfo.getUserName())) {
+            throw new UnauthorizedException();
+        }
```
anetteOlli, line 65: Nice

search-commons/src/main/java/no/unit/nva/search2/resource/ResourceClient.java
```diff
@@ -27,7 +27,7 @@ public class ResourceClient extends OpenSearchClient<SwsResponse, ResourceSearch
 
     private final UserSettingsClient userSettingsClient;
     @SuppressWarnings("PMD.DoNotUseThreads")
-    private static final ExecutorService executorService = Executors.newFixedThreadPool(5);
+    private static final ExecutorService executorService = Executors.newFixedThreadPool(3);
```
anetteOlli, line 30: Is the assignee the reason for decreasing the thread pool? In that case it is tempting to force the frontend to make two queries on their own and merge them. As this I would reckon causes all searches to a bit slower.

anetteOlli (approved):

search-resources-api/src/main/java/no/unit/nva/search2/SearchTicketAuthHandler.java
StigNorland: I do check for access in QueryBuilder as well, but it should be faster to check before invoking the builder.

search-commons/src/main/java/no/unit/nva/search2/resource/ResourceClient.java
StigNorland: To be honest, I am a bit confused as of why it is marked as a change, as I didn't change it now.

This code will never issue more than two async requests, so three should be enough...

PR #547, opened 2024-08-08
ScrollQueryBug + contributors + existQuery + improved parentRef

StigNorland:


search-commons/src/main/java/no/unit/nva/search/common/builder/OpensearchQueryExists.java
```diff
@@ -0,0 +1,68 @@
...
+import static no.unit.nva.search.common.constant.Words.KEYWORD_FALSE;
+import static org.opensearch.index.query.QueryBuilders.boolQuery;
+import static org.opensearch.index.query.QueryBuilders.existsQuery;
+
+/**
+ * @author Stig Norland
+ */
+public class OpensearchQueryExists<K extends Enum<K> & ParameterKey> extends OpensearchQuery<K> {
+
+    public static final String EXISTS_ANY = "ExistsAny-";
```
truhacevkir, line 23: So this one will be used when API user asks if some fields with some values exist/are present on a resource not matter value?

truhacevkir (approved):

search-commons/src/main/java/no/unit/nva/search/common/builder/OpensearchQueryExists.java
StigNorland: correct

PR #555, opened 2024-08-28
hotfix/ParameterKey-K

StigNorland:
In order for subQueries to work, ParameterKey have to have reference to instanceType
this is a prereq for parent child


LarsV123 (dismissed):
Looks good and the tests pass :-)

LarsV123 (approved):

PR #560, opened 2024-09-04
we need to add these two in order to get most of the queries right.

StigNorland:


search-commons/src/main/java/no/unit/nva/search/resource/ResourceParameter.java
```diff
@@ -426,31 +432,4 @@ public String errorMessage() {
...
-
-    private static boolean ignoreInvalidKey(ResourceParameter enumParameter) {
-        return enumParameter.ordinal() > IGNORE_PARAMETER_INDEX;
-    }
-
-    private static boolean isSearchField(ResourceParameter enumParameter) {
-        return enumParameter.ordinal() > IGNORE_PARAMETER_INDEX
-                && enumParameter.ordinal() < AGGREGATION.ordinal();
-    }
 }
```
StigNorland, line 435: Gjort av google-autoformatterer

search-commons/src/main/java/no/unit/nva/search/resource/ResourceParameter.java
```diff
@@ -266,7 +240,7 @@ public enum ResourceParameter implements ParameterKey<ResourceParameter> {
     TAGS_NOT(TEXT, NOT_ALL_OF, ENTITY_TAGS),
     TAGS_SHOULD(TEXT, ANY_OF, ENTITY_TAGS),
     TAGS_EXISTS(EXISTS, ANY_OF, ENTITY_TAGS),
-    TITLE(TEXT, ENTITY_DESCRIPTION_MAIN_TITLE, PI),
+    TITLE(TEXT, ALL_OF, ENTITY_DESCRIPTION_MAIN_TITLE, PI),
```
LarsV123, line 243: Is `ALL_OF` intended here?

LarsV123 (approved):
We should probably add the new parameters to the OpenAPI spec, but that's out of sync anyway so we could do it in a separate PR.

search-commons/src/main/java/no/unit/nva/search/resource/ResourceParameter.java
StigNorland: I removed the constructor, as it made no sense to have it, besides, I think it is better to declare `AND & OR` queries explicit.

PR #561, opened 2024-09-04
Removed SearchImportCandidatesHandler, tests & yaml definitions

StigNorland:


search-commons/src/main/java/no/unit/nva/search/importcandidate/ImportCandidateParameter.java
StigNorland, line 1: No need to test language features

LarsV123 (approved):

PR #565, opened 2024-09-10
Np 47323/np 47388 - verifiedOrNorwegian

StigNorland:


search-commons/src/main/java/no/unit/nva/search/resource/ContributorNodeFilter.java
```diff
@@ -0,0 +1,61 @@
...
+                .orElse((e) -> Boolean.TRUE);
+    }
+
+    private Boolean hasNoNorwegianBasedContributor(JsonNode element) {
+        return attempt(
+                        () ->
+                                !element.path(Words.AFFILIATIONS)
+                                        .findValues(Words.COUNTRY_CODE)
+                                        .contains(TextNode.valueOf(Words.NO)))
+                .orElse((e) -> Boolean.TRUE);
```
sondrev, line 61: Try to avoid using attempt() unless dealing with unknown errors. Here we know the only possible error: the field does not exist

Instead use JsonNode.path() with isMissingNode():

`public abstract JsonNode path(String fieldName) This method is similar to get(String), except that instead of returning null if no such value exists (due to this node not being an object, or object not having value for the specified field), a "missing node" (node that returns true for isMissingNode()) will be returned. This allows for convenient and safe chained access via path calls.`

search-commons/src/main/java/no/unit/nva/search/resource/ContributorNodeFilter.java
```diff
@@ -0,0 +1,67 @@
...
+                        () ->
+                                !element.path(AFFILIATIONS)
+                                        .findValues(COUNTRY_CODE)
+                                        .contains(TextNode.valueOf(NO)))
+                .orElse((e) -> Boolean.TRUE);
+    }
+
+    private Boolean deleteAfter(JsonNode element, int count) {
+        return attempt(() -> count < element.get(Constants.SEQUENCE).asInt()).orElse((e) -> Boolean.TRUE);
+    }
```
sondrev, line 66: Look at JsonNode.path() with isMissingNode(). 
Or: Optional.ofNullable(....get()).orElse() would be much clearer

search-commons/src/main/java/no/unit/nva/search/resource/ContributorNodeFilter.java
```diff
@@ -0,0 +1,67 @@
...
+    @Override
+    public JsonNode transform(JsonNode source) {
+        var target = source.at(CONTRIBUTORS_PATH_POINTER);
+        var elements = target.elements();
+        while (elements.hasNext()) {
+            var element = elements.next();
+            if (deleteAfter(element, FIRST_FIVE_CONTRIBUTORS)
+                    && isNotVerified(element)
+                    && hasNoNorwegianBasedContributor(element)) {
+                elements.remove();
```
sondrev, line 43: Instead of deleting everything after 5 - why not take the first 5? This would theoreticly be faster if there are thousands of contriibutors

search-commons/src/main/java/no/unit/nva/search/resource/ContributorNodeFilter.java
```diff
@@ -0,0 +1,67 @@
...
+
+    // (package private)
+    static final int FIRST_FIVE_CONTRIBUTORS = 5;
+    static final JsonPointer CONTRIBUTORS_PATH_POINTER = JsonPointer.compile(ENTITY_DESCRIPTION_CONTRIBUTORS_PATH);
+
+    @JacocoGenerated
+    ContributorNodeFilter() {}
+
+    public static ContributorNodeFilter verifiedOrNorwegian() {
+        return new ContributorNodeFilter();
```
sondrev, line 31: Why not just use the constructor and remove @JacocoGenerated ?

search-commons/src/main/java/no/unit/nva/search/resource/ContributorNodeFilter.java
```diff
@@ -0,0 +1,67 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.TextNode;
+
+import no.unit.nva.search.common.records.Midas;
+
+import nva.commons.core.JacocoGenerated;
+
+public class ContributorNodeFilter implements Midas {
+
+    // (package private)
```
sondrev, line 23: Is this comment necesary?

search-commons/src/main/java/no/unit/nva/search/common/records/Midas.java
```diff
@@ -0,0 +1,7 @@
+package no.unit.nva.search.common.records;
+
+import com.fasterxml.jackson.databind.JsonNode;
+
+public interface Midas {
+    JsonNode transform(JsonNode source);
```
sondrev, line 6: Whats a Midas? I havent heard about it before. Tried googling and still dont understand

sondrev (changes requested):

search-commons/src/main/java/no/unit/nva/search/resource/ContributorNodeFilter.java
sondrev: Or: Optional.ofNullable(....get()).orElse() w

search-commons/src/main/java/no/unit/nva/search/common/records/Midas.java
StigNorland: Midas transformed everything into gold...

search-commons/src/main/java/no/unit/nva/search/resource/ContributorNodeFilter.java
StigNorland: I have had question about this on qa's before, so I added it, to ensure there where no confusion about it's intentions

So, I wish it wasn't.

search-commons/src/main/java/no/unit/nva/search/resource/ContributorNodeFilter.java
StigNorland: All three statements has to be true, in order to remove an item.

search-commons/src/main/java/no/unit/nva/search/resource/ContributorNodeFilter.java
```diff
@@ -0,0 +1,57 @@
...
+
+    private Boolean isNotVerified(JsonNode element) {
+        return !VERIFIED.equals(element.path(IDENTITY + DOT + VERIFICATION_STATUS).asText());
+    }
+
+    private Boolean hasNoNorwegianBasedContributor(JsonNode element) {
+        return !element.path(AFFILIATIONS).findValues(COUNTRY_CODE).contains(TextNode.valueOf(NO));
+    }
+
+    private Boolean deleteAfter(JsonNode element, int count) {
```
sondrev, line 53: count is not really a counter, is it? How about sequenceNo

search-commons/src/main/java/no/unit/nva/search/resource/ContributorNodeFilter.java
```diff
@@ -0,0 +1,57 @@
...
+
+import com.fasterxml.jackson.core.JsonPointer;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.TextNode;
+
+import no.unit.nva.search.common.records.JsonNodeMutator;
+
+public final class ContributorNodeFilter implements JsonNodeMutator {
+
+    static final int FIRST_FIVE_CONTRIBUTORS = 5;
```
sondrev, line 20: Should be renamed to either "FIVE", because thats what the value is. Or "MINIMUM_INCLUDED_CONTRIBUTERS", because thats what the value controlls. Right now if we wanted to change it to 6 we would need to rename both value and name

search-commons/src/main/java/no/unit/nva/search/resource/ContributorNodeFilter.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Override
+    public JsonNode transform(JsonNode source) {
+        var target = source.at(CONTRIBUTORS_PATH_POINTER);
+        var elements = target.elements();
+        while (elements.hasNext()) {
+            var element = elements.next();
+            if (deleteAfter(element, FIRST_FIVE_CONTRIBUTORS)
+                    && isNotVerified(element)
+                    && hasNoNorwegianBasedContributor(element)) {
```
sondrev, line 38: Try shortering the transform() method by putting the 3 delete-checks in a shouldDeleteContributor() function

sondrev (approved):

PR #571, opened 2024-09-13
scientificValueOfChild

StigNorland:


LarsV123 (approved):
We should add a test case for this scenario, but this should work.

PR #577, opened 2024-09-23
added support for getMultiValueQueryStringParameters (fixes errors from repeated parameter keys)

StigNorland:


search-commons/src/main/java/no/unit/nva/search/common/ParameterValidator.java
```diff
@@ -177,15 +178,25 @@ public ParameterValidator<K, Q> fromRequestInfo(RequestInfo requestInfo) {
...
+    }
+
+    /**
+     * Adds testParameters from query.
+     *
+     * @apiNote This is intended to be used in runtime
+     */
+    public ParameterValidator<K, Q> fromMultiValueParameters(Map<String, List<String>> parameters) {
+        parameters.forEach((k, v) -> v.forEach(value -> setValue(k, value)));
+        return this;
```
sondrev, line 191: I cant see a test for this. I suspect it might put/set the latest value, and not use anythign else. Would be great with a test that proves that this approach works

sondrev (changes requested):

search-commons/src/main/java/no/unit/nva/search/common/ParameterValidator.java
StigNorland: Since fromMultiValueParameters is intended for runtime only, I updated a Handler test.

sondrev (approved):

PR #578, opened 2024-09-24
expand resoruces with promoted contributors

sondrev:


truhacevkir (commented):
Looks good, but I think that this expansion should be implemented as a part og expansion of publication in `ResourceExapsionServiceImpl` in `publication-api`. Otherwise we will end up with two expansion processes. 

Or is it an expansion that will be applied only once? Only when running KeyBasedBatchScan? Should it be applied when expanding document via 'default' flow, 
 i.e. when indexing single document?

indexing-handlers/src/main/java/no/unit/nva/indexing/utils/ResourceExpansion.java
```diff
@@ -0,0 +1,103 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.IntNode;
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+import java.util.Spliterators;
+import java.util.stream.StreamSupport;
+
+public class ResourceExpansion {
```
brinxmat, line 12: Here we affect a representation of Publication, which is the responsibility of publication-api. Adding this to search isn't acceptable.

indexing-handlers/src/main/java/no/unit/nva/indexing/utils/ResourceExpansion.java
```diff
@@ -0,0 +1,103 @@
...
+                            .sorted(ResourceExpansion::sortByPromoted)
+                            .limit(10)
+                            .toList();
+
+            ArrayNode arrayNode = new ArrayNode(JsonNodeFactory.instance);
+            arrayNode.addAll(contributorsPromoted);
+
+            ((ObjectNode) jsonNode.get(ENTITY_DESCRIPTION)).set(CONTRIBUTORS_PROMOTED, arrayNode);
+        }
+    }
```
truhacevkir, line 54: Can this method be shortened to something like:
```
    private static void expandPromotedContributors(JsonNode jsonNode) {
        var contributorNode = getContributorNode(jsonNode);
        if (!contributorNode.isMissingNode()) {
            var contributorList = convertToList(contributorNode);
            var contributorsPromoted = extractPromotedContributors(contributorList);
            injectPromotedContributorsNode(jsonNode, contributorsPromoted);
        }
    }
```

indexing-handlers/src/main/java/no/unit/nva/indexing/utils/ResourceExpansion.java
```diff
@@ -0,0 +1,103 @@
...
+            return false;
+        }
+        return verificationStatus.textValue().equals(VERIFIED);
+    }
+
+    public static JsonNode expand(JsonNode jsonNode) {
+        expandContributorCount(jsonNode);
+        expandPromotedContributors(jsonNode);
+        return jsonNode;
+    }
```
truhacevkir, line 102: Following method can also be void.

indexing-handlers/src/main/java/no/unit/nva/indexing/utils/ResourceExpansion.java
```diff
@@ -0,0 +1,103 @@
...
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.IntNode;
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+import java.util.Spliterators;
+import java.util.stream.StreamSupport;
+
+public class ResourceExpansion {
+
```
truhacevkir, line 13: private no-arg constructor

indexing-handlers/src/main/java/no/unit/nva/indexing/utils/ResourceExpansion.java
```diff
@@ -0,0 +1,103 @@
...
+        if (contributor1verification && !contributor2verification) {
+            return -1;
+        }
+
+        if (!contributor1verification && contributor2verification) {
+            return 1;
+        }
+
+        return 0;
+    }
```
truhacevkir, line 88: It seems like both methods can be implemented using comparator:

    private static int sortBySequence(JsonNode contributor1, JsonNode contributor2) {
        return Comparator.comparing((JsonNode node) -> node.get(SEQUENCE).isMissingNode())
                .thenComparingInt(node -> node.get(SEQUENCE).intValue())
                .compare(contributor1, contributor2);
    }

  ```
  private static int sortByPromoted(JsonNode contributor1, JsonNode contributor2) {
        return Comparator.comparing(ResourceExpansion::isContributorVerified)
                .reversed()
                .compare(contributor1, contributor2);
    }
```
Personally I think it is cleaner, but it is up to you :)

File: indexing-handlers/src/test/java/no/unit/nva/indexing/handlers/IndexResourceHandlerTest.java [hunk 6]
```diff
@@ -104,6 +116,66 @@ void shouldAddDocumentToIndexWhenResourceExistsInResourcesStorage() throws Excep
...
+    void shouldExpandDocumentIfItContainsContributors() throws IOException {
+
+        var json = createResourceJsonWithContributors();
+        var sampleResource =
+                createSampleResourceFromJson(SortableIdentifier.next(), RESOURCES, json);
+        URI resourceLocation = prepareEventStorageResourceFile(sampleResource);
+        InputStream input = createEventBridgeEvent(resourceLocation);
+        indexResourceHandler.handleRequest(input, output, context);
+        var indexedDocument =
+                indexingClient.getIndex(sampleResource.getIndexName()).stream().findFirst().get();
```
truhacevkir, line 129: `.orElseThrow()` instead of `get()`, it removes warning: `'Optional. get()' without 'isPresent()' check `

indexing-handlers/src/test/java/no/unit/nva/indexing/handlers/IndexResourceHandlerTest.java
```diff
@@ -104,6 +116,66 @@ void shouldAddDocumentToIndexWhenResourceExistsInResourcesStorage() throws Excep
...
+    private static ObjectNode createResourceJsonWithContributors() {
+        var root = new ObjectNode(JsonNodeFactory.instance);
+        var entityDescription = new ObjectNode(JsonNodeFactory.instance);
+        var contributorsList = new ArrayNode(JsonNodeFactory.instance);
+        contributorsList.add(createContributorJson(4, true));
+        contributorsList.add(createContributorJson(3, false));
+        contributorsList.add(createContributorJson(2, true));
+        contributorsList.add(createContributorJson(1, false));
+        entityDescription.put("contributors", contributorsList);
+        root.put("entityDescription", entityDescription);
```
truhacevkir, line 164: `put` is deprecated, should use `set()`

indexing-handlers/src/main/java/no/unit/nva/indexing/handlers/IndexResourceHandler.java
```diff
@@ -87,4 +89,9 @@ private IndexDocument fetchFileFromS3Bucket(UnixPath resourceRelativePath) {
         var resource = resourcesS3Driver.getFile(resourceRelativePath);
         return IndexDocument.fromJsonString(resource);
     }
+
+    private IndexDocument expandResource(IndexDocument indexDocument) {
+        ResourceExpansion.expand(indexDocument.resource());
+        return indexDocument;
+    }
```
truhacevkir, line 96: void method

StigNorland: then we end up with only 
`ResourceExpansion.expand(indexDocument.resource());`

StigNorland:
@sondrev  Denne kan vel lukkes nå?

PR #601, opened 2024-10-15
hotfix/NP-47861-sorting

StigNorland:
https://sikt.atlassian.net/browse/NP-47566
 https://sikt.atlassian.net/browse/NP-47861
 
 if search field identifier (sortable),is missing from sort keys, it is added last.

### DIFF
--- a/search-commons/src/main/java/no/unit/nva/search/common/AsType.java
+++ b/search-commons/src/main/java/no/unit/nva/search/common/AsType.java
@@ -114,7 +114,7 @@ public class AsType<K extends Enum<K> & ParameterKey<K>> {
 
     private static Stream<String> splitParts(String delimiter, String value) {
         return Arrays.stream(value.split(delimiter))
-                .filter(predicate -> !predicate.isEmpty())
+                .filter(predicate -> !predicate.isBlank())
                 .sequential();
     }
 
--- a/search-commons/src/main/java/no/unit/nva/search/common/AsType.java
+++ b/search-commons/src/main/java/no/unit/nva/search/common/AsType.java
@@ -1,5 +1,6 @@
 package no.unit.nva.search.common;
 
+import static nva.commons.core.StringUtils.EMPTY_STRING;
 import static nva.commons.core.attempt.Try.attempt;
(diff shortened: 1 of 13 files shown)


LarsV123 (approved):
Is this a duplicate of or fix for https://sikt.atlassian.net/browse/NP-47566 ?

LarsV123 (dismissed):

search-commons/src/main/java/no/unit/nva/search/common/AsType.java
```diff
@@ -134,6 +142,6 @@ public String asLowerCase() {
 
     @Override
     public String toString() {
-        return value;
+        return nonNull(value) ? value : EMPTY_STRING;
```
StigNorland, line 145: Part of the reason I created AsType, was to avoid null object.
Returning "" from toString when it is null or empty, doesn't seem to be a bad idea here....

search-commons/src/main/java/no/unit/nva/search/resource/ResourceSearchQuery.java
```diff
@@ -119,15 +120,19 @@ protected ResourceParameter keyFields() {
...
+            } else {
+                parameters().set(SORT, String.join(COMMA, sortString, IDENTIFIER));
+            }
         }
         return parameters().get(SORT);
     }
 
-    private static boolean missingIdentitySortKey(String sortString) {
-        return !sortString.contains(IDENTITY);
+    private static boolean missingIdentifier(String sortString) {
```
LarsV123, line 134: nitpick: Since this is a boolean, perhaps rename with `has` or `is` prefix? I suspect it might be more readable without this helper method too, since it's so short.

LarsV123 (approved):

search-commons/src/main/java/no/unit/nva/search/resource/ResourceSearchQuery.java
StigNorland: I tried with `hasMissingIdentifier`, but then the meaning changed....

PR #608, opened 2024-10-29
New response model

sondrev:


search-commons/src/main/java/no/unit/nva/search/resource/response/ResourceSearchResponse.java
```diff
@@ -0,0 +1,43 @@
...
+    private String mainLanguageAbstract;
+
+    @JsonProperty("description")
+    private String description;
+
+    @JsonProperty("alternativeTitles")
+    private List<String> alternativeTitles;
+
+    @JsonProperty("publicationDate")
+    private String publicationDate;
```
sondrev, line 33: @brinxmat This was in the document we looked at together, but not sure if we should have moved it to recordMetadata? Can you double check?

search-commons/src/main/java/no/unit/nva/search/resource/response/OtherIdentifiers.java
```diff
@@ -0,0 +1,23 @@
...
+    private List<String> cristin;
+
+    @JsonProperty("scopus")
+    private List<String> handle;
+
+    @JsonProperty("scopus")
+    private List<String> issn;
+
+    @JsonProperty("scopus")
+    private List<String> isbn;
```
brinxmat, line 21: Is it all the same?
If Scopus is everything,
Some data is lost!

search-commons/src/main/java/no/unit/nva/search/resource/response/ResourceSearchResponse.java
brinxmat: I guess it is the `{"year": "1232", "month": "12", "day": "02"}` publication date that is the intention

brinxmat (changes requested):
The classes can be records, which might save on annotations too

ketilaa (approved):

ketilaa:
Godkjente gitt at du vurderer å bruke records. La inn et lite spørsmål rundt publishing details.

search-commons/src/main/java/no/unit/nva/search/resource/response/PublishingDetails.java
```diff
@@ -0,0 +1,21 @@
+package no.unit.nva.search.resource.response;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+class PublishingDetails {
```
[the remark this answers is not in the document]
brinxmat: serialPublication := (journal, bookSeries, reportSeries, conferenceProceedingsSeries,…)

PR #610, opened 2024-11-25
Np 48081/tickets

StigNorland:


search-commons/src/main/java/no/unit/nva/search/common/QueryFilter.java
```diff
@@ -44,6 +49,6 @@ public void add(QueryBuilder builder) {
     public String toString() {
         return filters.values().stream()
                 .map(QueryBuilder::toString)
-                .collect(Collectors.joining(COMMA));
+                .collect(Collectors.joining(COMMA, "{", "}"));
```
LarsV123, line 52: This appears to "double-wrap" the values, is this intended?

Example output from this `toString` in a random test:
```
{{
  "terms" : {
    "status.keyword" : [
      "PUBLISHED",
      "PUBLISHED_METADATA"
    ],
    "boost" : 1.0,
    "_name" : "status"
  }
}}
```

search-commons/src/main/java/no/unit/nva/search/ticket/TicketAccessFilter.java
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    private boolean validateCustomManager(List<AccessRight> accessRights)
+            throws UnauthorizedException {
+        if (!accessRights.contains(MANAGE_CUSTOMERS)) {
+            throw new UnauthorizedException(
+                    USER_IS_NOT_ALLOWED_TO_SEARCH_FOR_TICKETS_NOT_OWNED_BY_THEMSELVES);
+        }
+        return true;
+    }
```
truhacevkir, line 255: Rename to hasAccessRight(MANAGE_CUSTOMER) or something similar?

search-commons/src/main/java/no/unit/nva/search/ticket/TicketAccessFilter.java
```diff
@@ -0,0 +1,276 @@
...
+     * <ul>
+     *   <lh>Business Rules.</lh>
+     *   <li>manage_doi -> DOI_REQUEST
+     *   <li>support -> GENERAL_SUPPORT_CASE
+     *   <li>manage_publishing_requests -> PUBLISHING_REQUEST
+     *   <li>manage_customers -> DOI_REQUEST, GENERAL_SUPPORT_CASE, PUBLISHING_REQUEST
+     *   <li>is_owner -> DOI_REQUEST, GENERAL_SUPPORT_CASE, PUBLISHING_REQUEST
+     * </ul>
+     */
+    private Set<TicketType> applyBusinessRules(List<AccessRight> accessRights)
```
truhacevkir, line 221: - Method name should describe what method does -> it returns tickets types user has right to search based on access rights. Rename it to something simple as getAllowedTicketTypes(accessRights).

- Should this filter throw exception? Or return empty ticket list when no access rights to search tickets with "not allowed" params?

search-commons/src/test/resources/resource_urls.json
```diff
@@ -1,4 +1,5 @@
 {
+  "query=Metoder+for+måle+vannføring+i+umålte+vassdrag": 0,
```
truhacevkir, line 2: This is not good test. I think we want a test that is named something like:

`shouldReturnTicketsUserHasAccessRightsToViewWhenSearchingForAllTickets()`

and 

`shouldNotReturnTicketsUserHasNoAccessRightsToViewWhenSearchingForAllTickets()`
 
and similar test for aggregations.

search-commons/src/main/java/no/unit/nva/search/ticket/TicketAccessFilter.java
```diff
@@ -0,0 +1,276 @@
...
+        }
+        if (accessRights.contains(AccessRight.SUPPORT) || isOwner || isCustomManager) {
+            allowed.add(TicketType.GENERAL_SUPPORT_CASE);
+        }
+        if (accessRights.contains(MANAGE_PUBLISHING_REQUESTS) || isOwner || isCustomManager) {
+            allowed.add(TicketType.PUBLISHING_REQUEST);
+        }
+        if (allowed.isEmpty()) {
+            throw new UnauthorizedException(USER_HAS_NO_ACCESS_TO_SEARCH_FOR_ANY_TICKET_TYPES);
+        }
```
truhacevkir, line 239: Maybe common check can be extracted in each if statement -> `doSomething(ACCESS_RIGHT)`

truhacevkir (commented):
Looks good, some comments. This is important functionality that user is not getting served something he/she has no rights to see, even if it is a aggregation number/counter. I think we want to documentate it with a proper test.

search-commons/src/main/java/no/unit/nva/search/common/QueryFilter.java
StigNorland: it is a simple way to resolve output problems in aws-log, replaced with '[' ']'

truhacevkir:
I would deploy it to dev and see if it fixes issues.

search-commons/src/test/resources/resource_urls.json
StigNorland: Added this test to prove that the error in test/dev, doesn't have anything to do with the code.....
https://sikt.atlassian.net/browse/NP-48140

search-commons/src/main/java/no/unit/nva/search/ticket/TicketAccessFilter.java
StigNorland: This procedure is the core of the accesslogic to Tickets.
It needs to be all there, no obscuring function that hides conditions and values.
It took me a very long time to get an overview of what rules were in play.

That is the reason for doing it this way.

search-commons/src/main/java/no/unit/nva/search/ticket/TicketAccessFilter.java
StigNorland: `var isOwner = searchAsTicketOwner() && validateOwner(currentUser);`

vs

`
        var isOwner = false; 
        if (searchAsTicketOwner()){
            validateOwner(currentUser);
            isOwner = true;
        }
`

StigNorland: When appling accessrights, and you ask for permissions you do not have, we do not continue execution, but throws `UnauthorizedException`

search-commons/src/main/java/no/unit/nva/search/ticket/TicketAccessFilter.java
StigNorland: The reason I renamed it, was that access right validation was handled here and there in this class.
Now it should be in only one place....

brinxmat: I feel like the response here is tangential to what was being asked — am I missing something?

search-commons/src/main/java/no/unit/nva/search/ticket/TicketAccessFilter.java
brinxmat: Again, I feel like I am missing something

ketilaa: It is the naming that is the issue here:

`private boolean validateCustomManager(List<AccessRight> accessRights)`

`customerManager` does not make sense here. `hasAccessToManageCustomers` is something that actually makes sense, and to me it triggers another question. Why would access to manage customers give users access to see other users tickets in the first place? I question why we do this check at all.

PR #622, opened 2024-12-23
boilerplate new index

sondrev:


indexing-handlers/src/main/resources/resource_20241201_mappings.json
```diff
@@ -0,0 +1,5 @@
+{
+  "properties": {
+
```
sondrev, line 3: Add fields during https://sikt.atlassian.net/browse/NP-48279

StigNorland (approved):

PR #627, opened 2025-01-09
added module for generating jaxb-annotated classes

ketilaa:
based on xsd schema files


oai-pmh/build.gradle
```diff
@@ -0,0 +1,15 @@
...
+    defaultPackage.set('org.openarchives.oai.v2.generated')
+}
+
+tasks.withType(Checkstyle).configureEach {
+    exclude { '**/generated/*' }
+}
+
+tasks.withType(Pmd).configureEach {
+    exclude { '**/generated/*' }
+}
```
brinxmat, line 15: Is there a generic "verify" task (that includes code coverage) that can be used to ignore more generally?

ketilaa: Best way I found that actually works is to make the check task depend on nothing:

`tasks.named("check").configure {
    dependsOn = []
}`

Committed!

brinxmat (approved):

PR #654, opened 2025-04-03
Code for new endpoint for searching within a users own resources

ketilaa:
Have not taken the cost to refactor the query builders to make them really testable, hence the split tests with mocks and container-based tests in search-commons.

Updates to openapi spec and cloud formation template will be handled in different PR's.


search-commons/src/test/java/no/unit/nva/search/resource/LegacyMutatorTest.java
```diff
@@ -0,0 +1,39 @@
+package no.unit.nva.search.resource;
```
ketilaa, line 1: Test coverage drops in search-commons without these tests

search-commons/src/test/java/no/unit/nva/search/resource/ResourceClientTest.java
```diff
@@ -1153,7 +1152,53 @@ void shouldFindBySeriesIssn(
     assertThat(actualIdentifiers, containsInAnyOrder(expectedIdentifiers));
   }
 
-  private static @NotNull Set<String> extractActualIdentifiersFromHits(PagedSearch pagedResponse) {
+  @Test
```
ketilaa, line 1155: Tests doing actual tests against container using more or less the same ResourceSearchQuery builder settings. Not great, but the ResourceSearchQuery needs refactoring to be really testable.

search-handlers/src/test/java/no/unit/nva/search/UserBasedResourceSearchHandlerTest.java
```diff
@@ -0,0 +1,139 @@
+package no.unit.nva.search;
```
ketilaa, line 1: Tests that mocks the actual OpenSearch client verifying that the generated query has the expected paramters and filters.

search-handlers/src/test/java/no/unit/nva/search/UserBasedResourceSearchHandlerTest.java
```diff
@@ -0,0 +1,137 @@
...
+            .withPageParameter(ResourceParameter.SIZE, "15")
+            .withPageParameter(ResourceParameter.SORT, "relevance,identifier")
+            .withSearchParameter(ResourceParameter.AGGREGATION, "none")
+            .withNamedFilterQuery(
+                "owner", new TermsQueryBuilderExpectation("resourceOwner.owner.keyword", username))
+            .build();
+    verify(resourceClient, Mockito.times(1)).doSearch(argThat(matcher));
+  }
+
+  private static @NotNull SwsResponse emptySwsResponse() {
```
truhacevkir, line 128: `@NotNull`

search-testing/src/main/java/no/unit/nva/search/testing/common/ResourceSearchQueryMatcher.java
```diff
@@ -0,0 +1,92 @@
...
+      Map<ResourceParameter, String> expectedParameters) {
+    var containsAll = true;
+    for (Map.Entry<ResourceParameter, String> entry : expectedParameters.entrySet()) {
+      if (!actualParameters.contains(entry)) {
+        containsAll = false;
+        break;
+      }
+    }
+    return containsAll;
+  }
```
truhacevkir, line 57: We can maybe make these two methods cleaner:

```
  private boolean hasNamedTermQuery(QueryFilter filters) {
    return namedFilterQueries.entrySet().stream()
        .allMatch(entry -> hasTermsQuery(filters, entry));
  }

  private static boolean hasTermsQuery(
      QueryFilter filters, Entry<String, TermsQueryBuilderExpectation> entry) {
    return filters.hasTermsQuery(entry.getKey(), entry.getValue().fieldName(), entry.getValue().values());
  }

  private boolean hasParameters(
      Set<Entry<ResourceParameter, String>> actualParameters,
      Map<ResourceParameter, String> expectedParameters) {
    return actualParameters.containsAll(expectedParameters.entrySet());
  }
```

truhacevkir (dismissed):

search-commons/src/main/java/no/unit/nva/search/common/QueryFilter.java
```diff
@@ -43,6 +47,15 @@ public int size() {
     return filters.size();
   }
 
+  public boolean hasTermsQuery(String queryName, String fieldName, Object... values) {
+    var query = this.filters.get(queryName);
+    if (nonNull(query) && query instanceof TermsQueryBuilder termsQueryBuilder) {
+      return termsQueryBuilder.fieldName().equals(fieldName)
+          && new HashSet<>(termsQueryBuilder.values()).containsAll(List.of(values));
+    }
```
brinxmat, line 55: This returns false if I enter nonsense like "heart=emoji"?

search-commons/src/test/java/no/unit/nva/search/resource/ResourceClientTest.java
```diff
@@ -1153,7 +1152,53 @@ void shouldFindBySeriesIssn(
...
+          "018c443840f1-b14a5947-2eae-4d67-9d95-c2e591059942"
+        };
+    var actualIdentifiers = extractActualIdentifiersFromHits(pagedResponse);
+
+    assertThat(actualIdentifiers, containsInAnyOrder(expectedIdentifiers));
+  }
+
+  @Test
+  void shouldNotBeAbleToBypassOwnResourcesOnlyFilter() throws BadRequestException {
+    var response =
```
brinxmat, line 1184: There's an argument that Forbidden and Bad Request aren't relevant for search. We should document this choice.

search-commons/src/main/java/no/unit/nva/search/common/QueryFilter.java
ketilaa: These are term-based queries applied programmatically as static filters. The method is only present for tests to be able to inspect and assert on presence of such filters.

search-commons/src/test/java/no/unit/nva/search/resource/ResourceClientTest.java
ketilaa: Ok, the appearance of BadRequestException in this test is due to a "feature" that validates query parameters and "helps" the client in understanding that it is using a non-supported parameter in the query. It could be an approach to just ignore unknown parameters, but then clients may mis-interprete the search result for something it is not. I will not change this now, but rather make sure it is documented when I update the open api specification for the endpoints. It may be documented for the existing endpoints already.

When it comes to Forbidden which is actually returned if the user/client making the request does not have access to create publications in the solution, I may be tempted to agree that an empty result set might be just as good. But it is clearer to actually tell the client that it is not authorized to make the request. I will not change this, but make sure it is documented in the next PR with the infrastructure-as-code that comes with this change.

PR #655, opened 2025-04-04
creating allScientificValues query param

truhacevkir:
Adding query param that "fixes" search results returned in "Tellende kategorier i ikke tellende kanaler". Right now this "retteliste" returns publications where two channels are present, but only one of the channels is not "tellende". Publication like this should not be returned in that list, cause they are valid nvi candidates and should not be returned in retteliste.


search-commons/src/test/java/no/unit/nva/search/resource/ResourceClientAllScientificValuesTest.java
```diff
@@ -0,0 +1,295 @@
...
+  private static void createIndexAndIndexDocument(String json)
+      throws IOException, InterruptedException {
+    indexingClient.createIndex(indexName, RESOURCE_MAPPINGS.asMap(), RESOURCE_SETTINGS.asMap());
+    indexingClient.refreshIndex(indexName);
+    var document =
+        new IndexDocument(
+            new EventConsumptionAttributes(indexName, SortableIdentifier.next()),
+            JsonUtils.dtoObjectMapper.readTree(json));
+    indexingClient.addDocumentToIndex(document);
+    Thread.sleep(DELAY_AFTER_INDEXING);
```
LarsV123, line 293: This is probably not necessary and will be very slow. Can we do the same "force refresh" trick that NVI uses? Not sure, since the OpenSearch client is different there.

See `no.sikt.nva.nvi.index.aws.OpenSearchClientTest#refreshIndex` in `nva-nvi`.

search-commons/src/test/java/no/unit/nva/search/resource/ResourceClientAllScientificValuesTest.java
LarsV123, line 1: You could perhaps make these tests more readable by splitting it into two parameterized tests? I.e. one with all examples where we expect a result and a separate test with all examples that should not be results. Using `argumentSet` and a `@MethodSource` lets you give each example a name.

Could potentially also merge those into a single index insert (for efficiency), but better to avoid it I think.

search-commons/src/test/java/no/unit/nva/search/resource/ResourceClientAllScientificValuesTest.java
truhacevkir: Forgot to remove this one, nice.

search-commons/src/test/java/no/unit/nva/search/resource/ResourceClientAllScientificValuesTest.java
truhacevkir: We do not want to test with dataset. I think that in search it is extra important to see what is lying in index and what fields we search for.

torbjokv (dismissed):

PR #656, opened 2025-04-07
Fixing dataset tests

truhacevkir:
When introducing proper tests running against existing indexes, it became evident that the order in which tests are executed affects the outcomes. To decouple "data set" tests from other test classes where we: initiate index -> run queries -> delete index, I implemented the following changes:

Test Class Organization:

Large test classes that run "data set" based tests now point to specific indexes such as `/resources` and `/tickets`. (as before)
All other test classes create a random index and execute queries against that particular index.
Dynamic Index Management:

During search operations, we pass the index name as a parameter instead of hardcoding it within the client.

**Main Change in Production Code:**
Instead of this: `query.doSearch(opensearchClient)`
We now do this: `query.doSearch(opensearchClient, indexName)`


search-commons/src/main/java/no/unit/nva/search/common/OpenSearchClient.java
```diff
@@ -56,7 +56,7 @@ public OpenSearchClient(HttpClient httpClient, CachedJwtProvider jwtProvider) {
     this.jwtProvider = jwtProvider;
   }
 
-  public R doSearch(Q query) {
+  public R doSearch(Q query, String indexName) {
```
truhacevkir, line 59: This is the main change.

oai-pmh-handler/src/test/java/no/sikt/nva/oai/pmh/handler/OaiPmhHandlerTest.java
ketilaa, line 1: Maybe even oai-pmh-handler should use a random index name?

search-commons/src/main/java/no/unit/nva/search/ticket/TicketSearchQuery.java
```diff
@@ -141,8 +140,8 @@ protected String[] include() {
   }
 
   @Override
-  public URI openSearchUri() {
-    return fromUri(infrastructureApiUri).addChild(TICKETS, SEARCH).getUri();
+  public URI openSearchUri(String indexName) {
+    return fromUri(infrastructureApiUri).addChild(indexName).addChild(SEARCH).getUri();
```
ketilaa, line 144: Did you miss the TICKETS part here?

ketilaa (approved):
Still feel that the index name knowledge is spread all around, and not simply injected for tests. But to fix it, more changes are needed. I think this is ok for now.

search-commons/src/main/java/no/unit/nva/search/ticket/TicketSearchQuery.java
truhacevkir: indexName provided in input is "tickets" now :)

PR #671, opened 2025-05-21
Attempt at targeting issues with unstable tests on AWS CodeBuild

ketilaa:
- now uses refresh index rather than static wait before using index in tests
- parallellized creation and population of indexes


search-commons/src/test/java/no/unit/nva/common/Containers.java
```diff
@@ -49,25 +53,50 @@ public class Containers {
   public static void setup() throws IOException, InterruptedException {
     container.withEnv("indices.query.bool.max_clause_count", "2048").start();
 
+    container.waitingFor(Wait.forListeningPort().withStartupTimeout(Duration.ofSeconds(60)));
```
ketilaa, line 56: Make sure container is ready and listening

brinxmat (approved):
Give it a go!

PR #693, opened 2025-08-07
Hotfix identifier handling get record

ketilaa:
It turned out that our search implementation did not support search by publication id (uri), so I have to get the local identifier when searching.

### DIFF
--- a/buildspec.yaml
+++ b/buildspec.yaml
@@ -31,6 +31,7 @@ phases:
         fi
   build:
     commands:
+      - ./gradlew test
       - sam build
   post_build:
     commands:


oai-pmh-handler/src/main/java/no/sikt/nva/oai/pmh/handler/oaipmh/handlers/GetRecordRequestHandler.java
```diff
@@ -16,11 +19,17 @@
...
   public GetRecordRequestHandler(
-      ResourceRepository resourceRepository, RecordTransformer recordTransformer) {
+      ResourceRepository resourceRepository,
+      RecordTransformer recordTransformer,
+      URI identifierBaseUri) {
     this.resourceRepository = resourceRepository;
     this.recordTransformer = recordTransformer;
+    this.localIdPattern =
+        Pattern.compile(
+            "^" + identifierBaseUri.toString().replaceAll("\\.", "\\\\.") + "/([^/]+)$");
```
brinxmat, line 32: Can you not just do UriWrapper::lastPathElement?

ketilaa: We could, but we would act on every weird domain uri out there. I think we should make sure the identifier points at where we expect before doing a search.

brinxmat (approved):

PR #694, opened 2025-08-11
Np 49380 other identifiers

ketilaa:
Added additional identifiers:
* NVA DOI
* reference DOI
* ISBN's
* ISSN's
* Handle
* Cristin identifier
* Scopus identifier

Refactored test data generation.

### DIFF
--- a/buildspec.yaml
+++ b/buildspec.yaml
@@ -9,6 +9,12 @@ cache:
   paths:
     - '/root/.gradle/caches/**/*'
 
+reports:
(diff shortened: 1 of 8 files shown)


ketilaa:
Probably our simplified model from search should have only one DOI. As it was, the field was located in `publishingDetails` and thereby references the DOI from where it was originally published. Hopefully, an NVA DOI is never created when another DOI already exists, so the OAI-PMH protocol with only expose one.

PR #712, opened 2025-09-05
feat: Add indexDocumentCreatedAt timestamp

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-49736

Adds a timestamp to all index documents to indicate when it was processed. This is intended to make it easier to clean up orphaned index documents later and to check that all documents have been processed after we redistribute documents between shards.


truhacevkir (dismissed):
Nice. How we plan to delete "orphans" based in this time stamp?

PR #714, opened 2025-09-10
[np-49735] fix: Distribute documents between shards

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-49735

This adds logic around how documents are inserted into OpenSearch indexes in order to ensure that:
1. Documents get distributed between all available shards fairly
2. Documents with a parent/child relationship are placed in the same shard

Changes:

- Removes static shard that routed all documents to the same shard
- Calculates target shard by selecting hash string from document content
  - Documents with a parent referenced use the `identifier`  of the parent document
  - Otherwise, use the document `identifier` or `id`
  - Fallback to using the entire document string as hash string (default OpenSearch behavior) if no identifiers are found


search-commons/src/test/java/no/unit/nva/indexingclient/utils/ShardRoutingUtils.java
```diff
@@ -0,0 +1,101 @@
...
+import no.unit.nva.indexingclient.models.IndexDocument;
+
+public class ShardRoutingUtils {
+  public static final String ID_FIELD = "id";
+  public static final String IDENTIFIER_FIELD = "identifier";
+  public static final String JOIN_FIELD = "joinField";
+  public static final String JOIN_FIELD_PARENT = "parent";
+  public static final String JOIN_FIELD_NAME = "name";
+  public static final String PART_OF = "partOf";
+  public static final String PLACEHOLDER_FIELD = "placeholder";
```
brinxmat, line 19: Should these be public?

search-commons/src/test/java/no/unit/nva/indexingclient/utils/ShardRoutingUtils.java
```diff
@@ -0,0 +1,101 @@
...
+    return resource;
+  }
+
+  public static String createResourceAndGetRoutingKey(String identifier) {
+    var resource = createDocumentNode();
+    resource.put(IDENTIFIER_FIELD, identifier);
+    return ShardRoutingService.calculateRoutingKey(resource);
+  }
+
+  public static ObjectNode createAnthologyDocument(String anthologyId) {
```
brinxmat, line 73: Is there a reason to not use the constants from this point on?

search-commons/src/test/java/no/unit/nva/indexingclient/ShardRoutingServiceTest.java
```diff
@@ -0,0 +1,291 @@
...
+    for (int i = 0; i < totalDocuments; i++) {
+      var publicationDocument = createStandalonePublicationDocument("publication-" + i);
+      var indexDocument = toIndexDocument(publicationDocument, RESOURCES_INDEX);
+
+      var indexRequest = indexDocument.toIndexRequest();
+      routingKeys.add(indexRequest.routing());
+    }
+
+    // Then routing keys should match publication identifiers
+    for (int i = 0; i < totalDocuments; i++) {
```
brinxmat, line 244: Here and elsewhere, you can use `var` rather than ìnt`

brinxmat (approved):

search-commons/src/test/java/no/unit/nva/indexingclient/utils/ShardRoutingUtils.java
LarsV123: Laziness? It should probably be refactored, but the reason was that I needed to use them in both test suites

LarsV123:
Accidentally dismissed approvals, so just merging it. Revert it if there are any issues.

PR #768, opened 2026-03-18
fix: Limit relevant fields in default wildcard search

LarsV123:
NOTE: This is a possible fix to address the issue described in https://sikt.atlassian.net/browse/NP-50823. Because it changes default behavior significantly, it will likely break something unexpected and should be tested carefully.

## Problem

The `CROSS_FIELDS` multi-match query in `searchAllWithBoostsQuery` used `"*"` (wildcard) for the default field set, which OpenSearch expands to every field in the index (~433 text fields in prod). Combined with `Operator.AND`, this creates `tokens × fields` boolean clauses.

The `.limit(7)` caps space-separated words, but the standard tokenizer also splits on hyphens — so a query like *"Genome-wide association meta-analysis..."* (7 words) produces 9+ tokens. With ~433 fields, that's ~3900–4300+ clauses, exceeding the `maxClauseCount` of 4096. Production has more dynamically-mapped fields than test, which is why the same queries work in test but fail in prod.

## Fix

Replaced the `"*"` wildcard with an explicit `DEFAULT_SEARCH_ALL_FIELDS` map containing 15 curated fields that are meaningful for free-text publication search:

- Title (boosted to PI), abstract, tags
- Identifiers (identifier, publisher ID, DOI)
- Contributor names
- Journal/publisher names
- Affiliation labels (Norwegian, English, Nynorsk — text fields for proper analyzer support)
- Funding identifiers

**Clause count: 15 fields × ~10 tokens = ~150 clauses** (vs 4000+ before), well within the 4096 limit — even with hyphenated queries.

When users explicitly specify the `fields` parameter (`NODES_SEARCHED`), the behavior is unchanged — only the default "search all" case is affected.

## Alternative: `copy_to` field

If the curated field list turns out to be too restrictive, a more robust long-term option is to add a `copy_to` directive in the index mapping that copies all searchable fields into a single combined text field (e.g. `_search_all`). The multi-match query would then target just 1 field instead of N, eliminating the clause explosion entirely while still supporting true all-field search. The trade-off is that it requires a mapping change and full re-index.


search-commons/src/main/java/no/unit/nva/search/resource/ResourceStreamBuilders.java
```diff
@@ -82,6 +95,25 @@ public class ResourceStreamBuilders {
...
+          Map.entry(jsonPath(ENTITY_DESCRIPTION, REFERENCE, DOI, KEYWORD), 1F),
+          Map.entry(jsonPath(DOI, KEYWORD), 1F),
+          Map.entry(jsonPath(ENTITY_CONTRIBUTORS, IDENTITY, NAME, KEYWORD), 1F),
+          Map.entry(jsonPath(ENTITY_PUBLICATION_CONTEXT, NAME, KEYWORD), 1F),
+          Map.entry(jsonPath(ENTITY_PUBLICATION_CONTEXT, PUBLISHER, NAME, KEYWORD), 1F),
+          Map.entry(jsonPath(CONTRIBUTORS_AFFILIATION_LABELS, BOKMAAL_CODE), 1F),
+          Map.entry(jsonPath(CONTRIBUTORS_AFFILIATION_LABELS, ENGLISH_CODE), 1F),
+          Map.entry(jsonPath(CONTRIBUTORS_AFFILIATION_LABELS, NYNORSK_CODE), 1F),
+          Map.entry(FUNDING_SOURCE_IDENTIFIER, 1F),
+          Map.entry(FUNDINGS_IDENTIFIER, 1F));
```
truhacevkir, line 115: This should be discussed with someone I think. It excludes many fields that are supported by free text search today.

LarsV123: Which fields?

truhacevkir: series, journals, projects and more.

joachimjorgensen: I'll call in a meeting with PO's where they define what fields they want searchable in the query.

PR #779, opened 2026-04-22
[np-51039] refactor: Extract and simplify reused validation

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-51039

Extracts and simplifies validation logic duplicated across `ResourceSearchQuery`, `TicketSearchQuery`, and `ImportCandidateSearchQuery` to the base class to make it easier to maintain.


torbjokv (approved):
nice!  mostly nitpick comments

PR #782, opened 2026-04-30
fix(oai-pmh-handler): strip XML-illegal control chars from marshalled output

torbjokv:
## Summary
External harvesters (Oria) saw HTTP 200 responses from OAI-PMH ListRecords that failed XML parsing, because contributor names, organization labels, publisher/journal names and identifiers were emitted with raw `\x01..\x1F` control bytes that XML 1.0 forbids. Sanitization is now done at the marshaller's writer boundary so every value flowing through JAXB is filtered, not a hand-curated set of fields.

- new `XmlCharFilteringWriter` wraps the `StringWriter` passed to the marshaller and drops characters outside the XML 1.0 `Char` whitelist (TAB, LF, CR, `0x20–0xFFFD`)
- `JaxbXmlSerializer.marshal` plugs the filter in, no other call sites change
- regression coverage: `firstPageHits.json` now carries `` bytes in a contributor name, so the existing list-records tests would fail XPath parsing without the fix
- unit tests cover all three `Writer.write` overloads

https://sikt.atlassian.net/browse/NP-51201

### DIFF
--- a/oai-pmh-handler/src/test/java/no/sikt/nva/oai/pmh/handler/OaiPmhHandlerTest.java
+++ b/oai-pmh-handler/src/test/java/no/sikt/nva/oai/pmh/handler/OaiPmhHandlerTest.java
@@ -785,6 +785,31 @@ class OaiPmhHandlerTest {
   }
 
   @Test
+  void shouldStripXmlIllegalControlCharactersFromContributorName()
+      throws IOException, JAXBException {
+    var pollutedContributorName = "Per Nordmann";
+    var sanitizedContributorName = "Per Nordmann";
+    var hit =
+        ResourceDocumentFactory.builder(
+                RESOURCE_ID, RESOURCE_TITLE, PUBLICATION_YEAR, PUBLICATION_MONTH, PUBLICATION_DAY)
+            .withTopLevelOrganization(NTNU_ID, NTNU_LABELS)
+            .withContributor(pollutedContributorName, null)
+            .academicArticle(new SerialChannelBuilder("Journal", JOURNAL_NAME))
+            .apply()
+            .build();
+    var inputStream = hitAndRequest(wrapHits(hit));
+
+    var gatewayResponse = invokeHandler(inputStream);
+
+    assertThat(gatewayResponse.getStatusCode(), is(equalTo(HTTP_OK)));
+    assertThat(gatewayResponse.getBody(), not(containsString("")));
+
+    var response = Input.fromString(gatewayResponse.getBody()).build();
+    var contributors = extractValuesFromDcElements("contributor", response);
+    assertThat(contributors, hasItem(sanitizedContributorName));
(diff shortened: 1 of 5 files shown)


oai-pmh-handler/src/test/resources/firstPageHits.json
LarsV123, line 1: Could we revert this and instead have an explicit test case in `OaiPmhHandlerTest`? For example like this:

```
  @Test
  void shouldStripXmlIllegalControlCharactersFromContributorName()
      throws IOException, JAXBException {
    var pollutedContributorName = "Per\u0001 Nord\u0001mann";
    var sanitizedContributorName = "Per Nordmann";
    var hit =
        ResourceDocumentFactory.builder(
                RESOURCE_ID, RESOURCE_TITLE, PUBLICATION_YEAR, PUBLICATION_MONTH, PUBLICATION_DAY)
            .withTopLevelOrganization(NTNU_ID, NTNU_LABELS)
            .withContributor(pollutedContributorName, null)
            .academicArticle(new SerialChannelBuilder("Journal", JOURNAL_NAME))
            .apply()
            .build();
    var inputStream = hitAndRequest(wrapHits(hit));

    var gatewayResponse = invokeHandler(inputStream);

    assertThat(gatewayResponse.getStatusCode(), is(equalTo(HTTP_OK)));
    assertThat(gatewayResponse.getBody(), not(containsString("\u0001")));

    var response = Input.fromString(gatewayResponse.getBody()).build();
    var contributors = extractValuesFromDcElements("contributor", response);
    assertThat(contributors, hasItem(sanitizedContributorName));
  }
```

LarsV123 (approved):

oai-pmh-handler/src/test/resources/firstPageHits.json
torbjokv: yup, makes sense

oai-pmh-handler/src/test/java/no/sikt/nva/oai/pmh/handler/OaiPmhHandlerTest.java
```diff
@@ -784,6 +784,31 @@ void shouldUseTopLevelOrganizationLabelAsRecordMetadataDcContributor()
     assertThat(contributors, hasItems(NTNU_LABEL_ENGLISH, UIB_LABEL_ENGLISH));
   }
 
+  @Test
+  void shouldStripXmlIllegalControlCharactersFromContributorName()
+      throws IOException, JAXBException {
+    var pollutedContributorName = "Per Nordmann";
```
joachimjorgensen, line 790: Is this polluted?

joachimjorgensen: Nvm, I just can't see it :D

PR #786, opened 2026-05-11
fix(bibtex): escape special chars outside math regions

torbjokv:
Previously `BibtexField.sanitize` bypassed all escaping whenever a value contained any LaTeX/MathML, leaving `%`, `#`, `&`, and unbalanced braces raw in the surrounding text — which breaks LaTeX rendering of the bibliography and can corrupt the `.bib` file (an unescaped `{` can swallow following entries).

- `sanitize` now splits on LaTeX/MathML spans and escapes only the non-math chunks, so math still renders and surrounding text is safe
- `MATH_OR_ML` regex tightened to also match full `<math>...</math>` elements (covers attributes like `style="font-size:120%"` that would otherwise be broken by escaping); dangling opening-tag detection kept as a fallback
- Deterministic tests added for `%`, `&`, `#`, and stray braces in mixed math-and-text values — covering both LaTeX (`\(...\)`) and MathML (`<math>...</math>`) variants
- Added explicit test for MathML element with escapable char in attribute, which justifies the full-element regex branch
- Parametrized the existing single-shot escape tests using `@CsvSource` / `@MethodSource` with `Named.of` for readable test reports

https://sikt.atlassian.net/browse/NP-51216


search-commons/src/test/java/no/unit/nva/search/common/bibtex/BibtexFieldPropertyTest.java
```diff
@@ -0,0 +1,176 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.util.List;
+import java.util.Random;
+import java.util.regex.Pattern;
+import java.util.stream.IntStream;
+import org.junit.jupiter.api.Test;
+
+class BibtexFieldPropertyTest {
```
joachimjorgensen: A bit confusing with a non deterministic test. What to do when it fails?

joachimjorgensen (approved):

PR #787, opened 2026-05-13
[np-51041] fix: Remove double-decoding and filter null values

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-51041

## Root causes
- API Gateway passes `null` values for bare query params like `?param` (no `=value`); these flowed into `getDecodedValue` and crashed with NPE in `String.replaceAll` / `URLDecoder.decode`.
- Malformed URL escapes (`%`, `%2`) reached `URLDecoder.decode` and threw `IllegalArgumentException` as a 5xx.
- Values were being URL-decoded twice on the way in (`getDecodedValue` then again in `QueryKeys.set`), which is what made the malformed-encoding case bubble out of multiple layers.

## Fix
- Filter nulls at the API Gateway boundary in `fromMultiValueParameters`.
- `getDecodedValue` catches malformed encodings, tracks the offending key, and `validate()` throws a typed `BadRequestException` instead of leaking the runtime exception.
- Remove the redundant second decode in `QueryKeys.set`. All call sites pass already-decoded or never-encoded values, verified across the 10 callers.

## Behavior change worth flagging
Values that were *double-URL-encoded* by clients (e.g. sending `%2525` expecting to search for `%20`) will now decode once instead of twice. No in-tree caller relied on this; the test suite (949 tests, coverage gate 97%/98.8%) passes unchanged.

[NP-51041]: https://sikt.atlassian.net/browse/NP-51041?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


brinxmat (approved):
Does this belong in commons in a sense?

PR #798, opened 2026-06-11
fix: Handle missing environment variables

LarsV123:
## Problem

Publishing a SnapStart version runs the function's full init phase to take the snapshot. StartBatchIndexingHandler's constructor loads the shared Constants class, whose static fields eagerly call ENVIRONMENT.readEnv("EVENT_BUS") and readEnv("TOPIC"), which throw IllegalStateException when the env var is missing. Those env vars  are only set on the two key-batch functions in template.yaml, not on StartBatchIndexingButton. This has been broken since September 2024 (commit 5ffc8f30): any manual invocation of that button lambda would have crashed at init. Nobody noticed because init only ran on invocation, and SnapStart now runs it at deploy time.   
  
 ## Changes

- Replaces `com.typesafe:config` library with `Environment.readEnvOpt` method
  - Relevant environment variables are read in `Constants` with `readEnvOpt(...)`  to handle missing variables
  - Handlers read via `Constants.VAR.orElseThrow()`, so missing variables throw exceptions only where they are explicitly needed
- Adds missing environment variables to `template.yaml`


batch-index/src/main/java/no/unit/nva/indexingclient/Constants.java
```diff
@@ -12,27 +11,29 @@
...
+      ENVIRONMENT.readEnvOpt("BATCH_INDEX_EVENT_BUS_NAME");
+  public static final Optional<String> EVENT_BUS = ENVIRONMENT.readEnvOpt("EVENT_BUS");
+  public static final Optional<String> RESOURCES_BUCKET =
+      ENVIRONMENT.readEnvOpt("PERSISTED_RESOURCES_BUCKET");
+  public static final Optional<String> KEY_BATCHES_BUCKET =
+      ENVIRONMENT.readEnvOpt("KEY_BATCHES_BUCKET");
+  public static final Optional<String> PERSISTED_RESOURCES_PATH =
+      ENVIRONMENT.readEnvOpt("PERSISTED_RESOURCES_PATH");
+  public static final Optional<String> TOPIC = ENVIRONMENT.readEnvOpt("TOPIC");
+  public static final String BATCH_INDEX_EVENT_TOPIC = "SearchService.Index.Batch";
```
torbjokv, line 30: private or package private?

torbjokv (approved):

batch-index/src/main/java/no/unit/nva/indexingclient/Constants.java
LarsV123: A couple of them could be, but most are used in sub-packages and need to be `public`. We might want to do something different entirely here and clean-up across the repo, but leaving them all `public` for now.

PR #802, opened 2026-06-16
test(search-commons): Verify BibTeX source-slim completeness and extend slimming to CSV

LarsV123:
Follow-up to #801 (now merged), which slimmed the OpenSearch `_source` for BibTeX exports to avoid a 413 from the search infrastructure (SWS). This adds the two pieces left on the table.

### 1. Completeness test for the BibTeX include list
`BibtexIncludedFieldsCompletenessTest` prunes each BibTeX real-data fixture down to `Constants.BIBTEX_INCLUDED_FIELDS` (emulating OpenSearch `_source` include filtering) and asserts the transformer produces byte-identical output to the full document, across all 7 fixtures.

Why: the field list is hand-maintained. If someone later makes the transformer read a field that is not in the list, the slimmed search response silently drops it and exports lose data with no test failure. This test fails loudly in that case.

### 2. Extend slimming to CSV
The inline CSV path on the search endpoints (`Accept: text/csv` on `/search/resources`) has the same unbounded `_source` problem BibTeX did. `ResourceSearchQuery.include()` never read `NODES_INCLUDED`, so the `NODES_INCLUDED` that `ExportResourceHandler` sets is actually inert for resources; the bulk CSV export only avoids 413 via its page-size-halving retry loop. Inline CSV has no such backstop.

This resolves the include set by media type, reusing `ResourceCsvTransformer.getJsonFields()` as the single source of truth for CSV fields (the same list the CSV export already intends to use).

### 3. Small refactor: `ResourceSourceFields`
The media-type-to-include-fields decision is extracted into a tiny helper. Adding the CSV branch inline tipped the PMD GodClass check on `ResourceSearchQuery` (an extra foreign-data access to `ResourceCsvTransformer`); moving the decision out keeps the query class under the threshold and makes the logic directly testable.

## Verification
- `./gradlew build` green (PMD, Checkstyle, JaCoCo coverage, dependency analysis).
- New query-layer test asserts CSV media type yields the CSV field set in the assembled OpenSearch request; #801's existing tests already cover BibTeX and the untouched default path.

## Note on tickets
This work and #801 both address the same SWS 413 on BibTeX export, tracked separately as NP-51328 (this) and NP-51352 (#801). They are duplicates; NP-51328 should be closed or linked.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/search-commons/src/main/java/no/unit/nva/search/resource/ResourceSourceFields.java
+++ b/search-commons/src/main/java/no/unit/nva/search/resource/ResourceSourceFields.java
@@ -1,0 +1,39 @@
+package no.unit.nva.search.resource;
+
+import static java.util.Objects.nonNull;
+import static no.unit.nva.constants.Defaults.BIBTEX_UTF_8;
+import static no.unit.nva.search.resource.Constants.BIBTEX_INCLUDED_FIELDS;
+import static nva.commons.apigateway.MediaType.CSV_UTF_8;
+
+import java.util.Collection;
+import no.unit.nva.search.common.csv.ResourceCsvTransformer;
+import nva.commons.apigateway.MediaType;
+
+/**
+ * Resolves the OpenSearch {@code _source} include fields for a resource search.
+ *
+ * <p>BibTeX and CSV exports return many hits, and the full document (contributor affiliations,
+ * files, funding) can exceed the search infrastructure response-size limit. For those media types
+ * the source is slimmed to the fields the transformer actually reads. JSON requests are unaffected
+ * and keep their configured include set.
+ */
+final class ResourceSourceFields {
+
+  private ResourceSourceFields() {}
+
+  static String[] forMediaType(MediaType mediaType, Collection<String> defaultIncludedFields) {
+    final Collection<String> includedFields;
+    if (matches(mediaType, BIBTEX_UTF_8)) {
+      includedFields = BIBTEX_INCLUDED_FIELDS;
+    } else if (matches(mediaType, CSV_UTF_8)) {
+      includedFields = ResourceCsvTransformer.getJsonFields();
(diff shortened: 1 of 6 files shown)


search-commons/src/main/java/no/unit/nva/search/resource/ResourceSourceFields.java
```diff
@@ -0,0 +1,39 @@
...
+ * and keep their configured include set.
+ */
+final class ResourceSourceFields {
+
+  private ResourceSourceFields() {}
+
+  static String[] forMediaType(MediaType mediaType, Collection<String> defaultIncludedFields) {
+    final Collection<String> includedFields;
+    if (matches(mediaType, BIBTEX_UTF_8)) {
+      includedFields = BIBTEX_INCLUDED_FIELDS;
```
torbjokv: flytt til ResourceBibTexTransformer?

search-commons/src/test/java/no/unit/nva/search/common/bibtex/BibtexIncludedFieldsCompletenessTest.java
```diff
@@ -0,0 +1,90 @@
...
+    JsonNode result;
+    if (node.isArray()) {
+      var filteredArray = MAPPER.createArrayNode();
+      node.forEach(element -> filteredArray.add(filter(element, pathTree)));
+      result = filteredArray;
+    } else if (node.isObject()) {
+      var filteredObject = MAPPER.createObjectNode();
+      node.fields()
+          .forEachRemaining(
+              entry -> {
```
torbjokv: tidy up the nested stuff here?

torbjokv (approved):

PR #809, opened 2026-06-17
Add json-ld export

brinxmat:
Adds schema.org JSON-LD as a new response format for the resource search endpoints, negotiated through the `Accept` header. Search hits are transformed into a schema.org `ItemList` where each hit becomes a typed item (ScholarlyArticle, Book, Chapter, Thesis, …) with contributors, publication context (journal/volume/issue, book/series, publisher), ISBN/ISSN, pages, DOI and keywords.

- Supports three `Accept` variants: `application/ld+json`, `application/vnd.schemaorg.ld+json`, and `application/ld+json; profile="https://schema.org"`
- Parses the full `Accept` header with `MediaTypeParser` and resolves the response type by client preference order, falling back to JSON
- Adds the `no.unit.nva.search.common.bibliography` package with the schema.org record model and transformers (NVA-type → schema.org-type mapping, contributor/role mapping, publication context, manifestations)
- Emits `X-Total-Count`, pagination `Link`, and a `rel="profile"` schema.org `Link` header for JSON-LD/schema.org responses
- Slims the OpenSearch `_source` to the fields the transformer reads (as already done for CSV/BibTeX) to stay within the response-size limit
- Registers the new media types on the resource search endpoints and documents them in `openapi.yaml`
- Restricts ticket and import-candidate search to `application/json` (they never meaningfully supported CSV/JSON-LD; CSV was disabled/broken)
- Bumps `nva-commons` to 2.8.3 (provides `MediaTypeParser` and `MediaTypes.SCHEMA_ORG`)

https://sikt.atlassian.net/browse/NP-51371


search-commons/src/main/java/no/unit/nva/search/common/bibliography/SchemaOrgBibliographyTransformer.java
```diff
@@ -0,0 +1,24 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
+import java.util.Collection;
+import no.unit.nva.commons.json.JsonUtils;
+
+public final class SchemaOrgBibliographyTransformer {
+
+  private static final ObjectMapper MAPPER =
+      JsonUtils.dtoObjectMapper.copy().disable(SerializationFeature.INDENT_OUTPUT);
```
torbjokv: why?

search-commons/src/test/java/no/unit/nva/search/common/records/HttpResponseFormatterPaginationHeadersTest.java
```diff
@@ -36,10 +36,32 @@ void shouldReturnEmptyHeadersWhenMediaTypeIsJson() {
...
+  void shouldEmitPaginationHeadersForJsonLd() {
     var headers = formatterFor(MediaTypes.APPLICATION_JSON_LD, 0, 10, 100).paginationHeaders();
 
-    assertThat(headers, anEmptyMap());
+    assertThat(headers, hasEntry(X_TOTAL_COUNT, "100"));
+    assertThat(headers, hasEntry(ACCESS_CONTROL_EXPOSE_HEADERS, EXPECTED_EXPOSED_HEADERS));
+  }
+
+  @Test
+  void shouldEmitProfileLinkForJsonLd() {
```
torbjokv, line 60: if you omit profile in request, it will default to our verbose json?

search-commons/src/main/java/no/unit/nva/constants/Words.java
```diff
@@ -138,6 +138,7 @@ public final class Words {
   public static final String STATUS = "status";
   public static final String SUFFIX = ")";
   public static final String TAGS = "tags";
+  public static final String APPLICATION_LD_JSON = "application/ld+json";
```
torbjokv: have a look at `static String[] forMediaType(MediaType mediaType, Collection<String> defaultIncludedFields)`

Could you add the needed json fields here to limit "Entity Too Large" issues?

torbjokv: (this comment is more or less randomly placed btw)

search-commons/src/main/java/no/unit/nva/search/common/bibliography/SchemaOrgBibliographyTransformer.java
brinxmat: No good reason, I guess more compact output.

search-commons/src/test/java/no/unit/nva/search/common/records/HttpResponseFormatterPaginationHeadersTest.java
brinxmat: This concern existed before this change, but I have now added a test that explicitly pins the default representation to JSON-UTF-8

search-commons/src/main/java/no/unit/nva/constants/Words.java
torbjokv: maybe we should restructure the pattern here so that included fields are enforced next time we add a new profile/format

brinxmat: I believe this is now addressed

search-commons/src/main/java/no/unit/nva/search/common/records/HttpResponseFormatter.java
```diff
@@ -226,6 +249,10 @@ public String toString() {
     if (BIBTEX_UTF_8.matches(this.mediaType)) {
       return toBibTexText();
     }
+    if (MediaTypes.APPLICATION_JSON_LD.matches(this.mediaType)
```
torbjokv, line 252: GET /search/tickets (and /import-candidates) with Accept: application/ld+json used to return normal JSON — now it returns mostly-empty items since they have no entityDescription. Should probably 406 instead. text/csv has the same issue here. DEFAULT_RESPONSE_MEDIA_TYPES shouldn't be shared — each endpoint needs to declare its own listSupportedMediaTypes().

search-commons/src/main/java/no/unit/nva/search/common/SearchQuery.java
```diff
@@ -204,9 +207,26 @@ public MediaType getMediaType() {
   }
 
   final void setMediaType(String mediaType) {
-    if (nonNull(mediaType) && mediaType.contains(Words.TEXT_CSV)) {
+    if (mediaType == null) {
+      this.mediaType = JSON_UTF_8;
+      return;
+    }
+    var parsed = MediaTypeParser.defaultParser().parse(mediaType).first();
```
torbjokv: parse() only accepts a single media type, so any Accept with more than one type (e.g. text/csv, application/json) is rejected with UNEXPECTED_LIST and parses nothing — first() is empty and we fall back to JSON_UTF_8. This happens on count alone, before q-values are considered. The UNEXPECTED_LIST violation is also swallowed (we only read first()), so it silently  returns JSON instead of 406. Should this use parseList() and pick by preferenceOrder()?

search-commons/src/main/java/no/unit/nva/search/common/bibliography/SchemaOrgItemList.java
```diff
@@ -0,0 +1,10 @@
+package no.unit.nva.search.common.bibliography;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.List;
+
+record SchemaOrgItemList(
+    @JsonProperty("@context") String context,
+    @JsonProperty("@type") String type,
+    int numberOfItems,
+    List<SchemaOrgItem> itemListElement) {}
```
torbjokv, line 10: schema.org says itemListElement is an unordered property, but its probably not here? consider wrapping SchemaOrgItem in ListItem that have position

search-commons/src/main/java/no/unit/nva/constants/Words.java
torbjokv: goodgood. this line is now unused

search-commons/src/main/java/no/unit/nva/search/common/SearchQuery.java
```diff
@@ -206,13 +207,26 @@ public MediaType getMediaType() {
   }
 
   final void setMediaType(String mediaType) {
-    if (nonNull(mediaType)
-        && mediaType.contains(APPLICATION_LD_JSON)
-        && mediaType.contains("profile=")) {
+    if (mediaType == null) {
```
torbjokv: isNull()

torbjokv (approved):

PR #812, opened 2026-06-30
NP-51404 Add NPI Subject Heading query param

joachimjorgensen:


docs/openapi.yaml
```diff
@@ -583,6 +584,8 @@ paths:
         - $ref: "#/components/parameters/licenseShouldParam"
         - $ref: "#/components/parameters/modifiedBeforeParam"
         - $ref: "#/components/parameters/modifiedSinceParam"
+        - $ref: "#/components/parameters/npiSubjectHeadingParam"
+        - $ref: "#/components/parameters/npiSubjectHeadingNotParam"
```
LarsV123: Should these be added to `/customer/resources`, `/user/resources`, and `/resources/export` too? These endpoints share the same `ResourceParameter`, so the filter should work there now and just isn't documented.

search-commons/src/main/java/no/unit/nva/search/resource/ResourceParameter.java
```diff
@@ -182,6 +183,10 @@ public enum ResourceParameter implements ParameterKey<ResourceParameter> {
   JOURNAL_SHOULD(FUZZY_KEYWORD, ANY_OF, ENTITY_DESCRIPTION_REFERENCE_JOURNAL),
   LICENSE(FUZZY_KEYWORD, ALL_OF, ASSOCIATED_ARTIFACTS_LICENSE),
   LICENSE_NOT(FUZZY_KEYWORD, NOT_ALL_OF, ASSOCIATED_ARTIFACTS_LICENSE),
+  NPI_SUBJECT_HEADING(KEYWORD, ANY_OF, ENTITY_DESCRIPTION_NPI_SUBJECT_HEADING),
+  NPI_SUBJECT_HEADING_NOT(KEYWORD, NOT_ANY_OF, ENTITY_DESCRIPTION_NPI_SUBJECT_HEADING),
+  NPI_SUBJECT_HEADING_SHOULD(FUZZY_KEYWORD, ANY_OF, ENTITY_DESCRIPTION_NPI_SUBJECT_HEADING),
```
LarsV123: This adds 4 parameters , but only 3 of them are included in the OpenAPI docs. Should `NPI_SUBJECT_HEADING_SHOULD` be documented too?

search-commons/src/main/java/no/unit/nva/search/resource/Constants.java
```diff
@@ -236,6 +237,8 @@ public final class Constants {
       jsonPath(ENTITY_DESCRIPTION, MAIN_TITLE);
   public static final String ENTITY_DESCRIPTION_MAIN_TITLE_KEYWORD =
       jsonPath(ENTITY_DESCRIPTION_MAIN_TITLE, KEYWORD);
+  public static final String ENTITY_DESCRIPTION_NPI_SUBJECT_HEADING =
+      jsonPath(ENTITY_DESCRIPTION, NPI_SUBJECT_HEADING);
```
LarsV123, line 241: Should this include `KEYWORD` like the `MAIN_TITLE` constant above it, so the queries target `.keyword`?

We don't have explicit mappings for this field, so it's dynamically mapped as both a `keyword` field (the exact value) and a normalized `text` field. Might not matter for numeric codes, but now a doc with NPI code `ABC123` gets mapped to `{text: "abc123", keyword: "ABC123"}` and wouldn't be found by a query for `ABC123`.

LarsV123 (approved):

docs/openapi.yaml
joachimjorgensen: Good catch!

search-commons/src/main/java/no/unit/nva/search/resource/ResourceParameter.java
joachimjorgensen: Added 👍

search-commons/src/main/java/no/unit/nva/search/resource/Constants.java
joachimjorgensen: Good to keep in mind, I didn't know that. Will add it even though the field is a numeric code 👍

PR #817, opened 2026-07-03
Add alternativeAbstracts again to SimplifiedMutator

joachimjorgensen:


torbjokv (commented):
The change itself is a clean re-add of the `alternativeTitles` pattern, but can potentially  re-introduce the exact crash we saw last deploy: `jsonNodeMapToMap` does `convertValue(..., Map<String,String>)`, so a single document with an array value (e.g. `"und": [...]`) throws `Cannot deserialize value of type String from Array value` — and since `toMutatedHits()` streams over all hits with `attempt().orElseThrow()`, one bad 
  document fails the whole search page (500), not just that field. Reindexing only hides this by normalizing the data; any non-conforming document re-triggers it there is stuff we forgot ti fix in the index.                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                                                                                                                                 
  Could we make search-api tolerant of the data shape instead of relying on the reindex? Suggestions:                                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                                                                                 
  1. Make `jsonNodeMapToMap` robust to value shape — coerce each value to text (join/first element for arrays) instead of blindly deserializing to `Map<String,String>`. This protects both `alternativeAbstracts` and `alternativeTitles`.                                                                                                                                                                                      
  2. Isolate per-hit in `toMutatedHits()` so one malformed document can't take down the entire page for *any* field.                                                                                                                                                                                                                                                                                                             
  3. Add a regression test with an array value (`"und": ["a","b"]`) asserting the transform doesn't throw — the current test only covers the string happy path.

brinxmat (approved):

LarsV123 (approved):

PR #820, opened 2026-07-14
[np-51095] fix: Limit wildcard search to curated field list

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-51095

This limits the wildcard search to a curated set of mapping fields so that we reduce the number of errors and increase the number of words allowed in a query. The old implementation capped queries at 7 words, this increases the cap to ~20 words.

This is a stop-gap measure and we can significantly improve on it when we change the index document model as planned.

Notable decisions:

- Long queries are truncated to avoid going past the clause limit
- If a query still manages to trigger a `TOO_MANY_NESTED_CLAUSES` exception from OpenSearch, we map this to a `400 BAD REQUEST` exception
- Some fields are deliberately left out from this curated set, which means tests that relied on these in free text search were changed. To support these fields with the current document structure, other fields would have to be removed from the list to compensate. These fields are no longer found by free-text search:
  - Organization labels (names), both top-level and sub-units
  - Publication channel IDs/identifiers for publisher and series (`publicationContext.publisher.id` etc)
  - Customer ID/identifier (`publisher.id`)


search-commons/src/main/resources/indices/resource/mappings.json
LarsV123, line 1: This is changed to match what we already have in the prod index, so no reindexing is required.

search-commons/src/main/java/no/unit/nva/search/resource/Constants.java
```diff
@@ -290,6 +292,16 @@ public final class Constants {
           jsonPath(ENTITY_PUBLICATION_INSTANCE, "corrigendumFor", KEYWORD),
           jsonPath(ENTITY_PUBLICATION_INSTANCE, "manifestations", ID, KEYWORD),
           jsonPath(ENTITY_PUBLICATION_INSTANCE, ID, KEYWORD));
+
+  /**
+   * List of language codes relevant for free-text search (Scandinavian languages, English, and
+   * undetermined).
+   */
+  private static final List<String> FREE_TEXT_LANGUAGE_CODES =
+      List.of("en", "eng", "nb", "nn", "no", "da", "sv", "und");
```
brinxmat, line 301: Why are only these languages are relevant?

brinxmat (approved):

search-commons/src/main/java/no/unit/nva/search/resource/Constants.java
LarsV123: It's a matter of priority. This sub-set was proposed based on analysis of prod data that showed we had very little data for other languages. See "Design decisions > Language selection" in https://sikt.atlassian.net/browse/NP-51095.

Because of the data structure in the search document, every language we support increases the number of clauses in the query and makes the original problem worse. If we change the data model for index documents and flatten it, then we can remove this and search across all languages again.

